PR #3, opened 2020-04-03
Add some error checking

io7m:
This adds some error checking to the `rebuild.sh` script.  If this
script is unwittingly executed without arguments, or executed on a
system that doesn't have docker installed, the results can be pretty
nasty. With the error checks in place, the script is prevented from
proceeding in a broken state.


hboutemy:
nice improvement, thank you

PR #4, opened 2020-04-04
buildspec for akka-actor_2.13

raboof:
Haven't looked into actually building/reproducing in this repo yet. Perhaps https://hub.docker.com/r/hseeberger/scala-sbt/tags could be a starting point, but Akka is peculiar in that it needs use jdk11 but also have jdk8 installed.

### DIFF
--- a/content/com/typesafe/akka/akka-actor_2.13-2.6.4.buildspec
+++ b/content/com/typesafe/akka/akka-actor_2.13-2.6.4.buildspec
@@ -1,0 +1,15 @@
+scalaBinaryVersion=2.13
+scalaVersion=2.13.1
+
+groupId=com.typesafe.akka
+artifactId=akka-actor_${scalaBinaryVersion}
+version=2.6.4
+
+gitRepo=https://github.com/akka/akka
+gitTag=v${version}
+
+jdk=11
+build-tool=sbt
+
+command=sbt ++${scalaVersion} -Dakka.genjavadoc.enabled=true akka-actor/package
+buildinfo=/akka-actor/target/scala-${scalaBinaryVersion}/akka-actor_${scalaBinaryVersion}-${version}.buildinfo


hboutemy:
thank you arnout
do you know a Docker image that contains sbt and JDK 11 to be used to easily rebuild?

hboutemy:
what does "use jdk 11 but also have jdk8 installed" mean?
2 different jdks are used during build?
I could not find the documentation on how to build this project

raboof:
> what does "use jdk 11 but also have jdk8 installed" mean? 2 different jdks are used during build?

Unfortunately yes: the jdk11 `-release` flag was not sufficient in all cases, and we actually need to build against the jdk8 standard library for part of the build.

The sbt command in the buildspec *should* be all you need.

The jdk11 docker image at https://hub.docker.com/r/hseeberger/scala-sbt/tags *looks* like a good start but I haven't dug in further yet.

hboutemy:
I just added in 7a82da531cc4100142bb1ff0a150fa7784118584 a new tool=sbt option in buildspec to prepare support for SBT
help wanted to implement support in `rebuild.sh`

hboutemy:
would https://repo.maven.apache.org/maven2/com/scalapenos/stamina_2.12/0.1.5/ be easier to reproduce?

raboof:
Possibly! I haven't kept such a close eye on that one.

hboutemy:
ok, I just created a sbt branch https://github.com/jvm-repo-rebuild/reproducible-central/tree/sbt
I added a buildspec for stamina, which seems a good candidate because it is small
then I updated rebuild.sh to try to rebuild: it does not fail, but it does not yet export anything beyond the temporary Docker container
I'll need your help to:
- make it export the result of the build to the local filesystem
- try to share local repository to avoid downloading everything at each run

raboof:
Had a quick look but have to go, added my WiP PR at #7

* make it export the result of the build to the local filesystem

Not found yet, haven't looked into why

* try to share local repository to avoid downloading everything at each run

Haven't looked into this yet, but probably sharing ~/.ivy2 or ~/.cache/coursier like you're probably already sharing ~/.m2 should do the trick

hboutemy:
looks like io/monix and com/typesafe/akka are publishing buildinfo files: @raboof did you know?

raboof:
I knew about Akka, but not about Monix. Nice.

PR #5, opened 2020-04-05
Add com.io7m.jade

io7m:
This adds a new com.io7m.jade buildspec for version 1.0.1.


hboutemy:
merged and fixed buildspec in dcee2f786a99566a1050913848ec579b5eccbf85 while merging result

thank you

io7m:
:+1:

Odd that the spec needed to refer to the `com.io7m.jade.documentation` module; that module only exists to aggregate documentation from the other modules. Is this a bug in the `buildinfo` plugin, or is there something weird about my project that causes this?

hboutemy:
not a bug, but a subtle consequence of the way buildinfo is generated during build: to be able to get the checksum for every artifact, it has to be run after everything, then in the last module, just before deploy

PR #7, opened 2020-04-11
Updates to stamina rebuilding

raboof:
Not quite there yet: doesn't correctly copy the buildinfo
(haven't dug into why)


hboutemy:
nice first step, thank you

PR #8, opened 2020-04-15
Share ivy2 and coursier caches across sbt builds

raboof:


raboof:
This still does not successfully place the buildinfo file in `content/com/scalapenos/stamina_2.13/buildcache/stamina/target/scala-2.13/stamina_2.13-0.1.5.buildinfo`:

The remaining problem is caused by the fact that `content/com/scalapenos/stamina_2.13/buildcache/stamina/` is created by the user 'outside' the container, which is different from the 'sbtuser' user that is used 'inside' the container.

I see 2 possible ways around this:
1) do the 'git clone' inside the container
2) use the same user 'inside' the container as 'outside'

'2' seems more elegant, but passing `-u "$(id -u):$(id -g)" -w /home/sbtuser/dev --env HOME=/home/sbtuser` was not sufficient: for some reason that leads to `file:/?/` URL's sbt. I haven't figured out yet what is missing there.

hboutemy:
superseded by #10

PR #10, opened 2020-05-01
Fix rebuilding the stamina root project with sbt

raboof:


hboutemy:
great work, thank you: I could reproduce on my computer (with obviously different timestamp in `date` property of buildinfo)
last step is to produce `.buildinfo.compare` file to have the reproducibility status visible in the `README.md` report
I'll hand-write a file for now, but to have multiple cases, it will be necessary to have that file generated

hboutemy:
notice while writing the .compare file: sbt buildinfo contains pom.xml in output, while Maven buildinfo does not
I think that I should add pom.xml in Maven output: this is useful now to match others, but also will be useful in the future with Maven 3.7.0, when pom.xml pushed to repository can be different from pom.xml used at build time...

raboof:
> I think that I should add pom.xml in Maven output: this is useful now to match others, but also will be useful in the future with Maven 3.7.0, when pom.xml pushed to repository can be different from pom.xml used at build time...

I agree! Also, otherwise, an attacker could sneak a malicious dependency into the pom.xml of a "verified reproducible" package.

PR #11, opened 2020-05-02
complete set of stamina 0.1.5 artifacts

raboof:
Stamina consists of multiple subprojects and is built for multiple minor Scala
versions. This completes all stamina 0.1.5 artifacts.


hboutemy:
yes, adding the complete set of artifacts makes sense

but way it is done by rebuilding stamina.git content for each 4 artifacts is not really the way we expect to rebuild a multi-module build. And in addition every build is done 3 times given the 3 Scala minor version (2.11, 2.12 and 2.13).
Then what is 1 unique release, becomes 3 (2.11, 2.12 and 2.13), each built 4 times (stamina, stamina-core, stamina-json and stamina-testkit) = 12 buildspecs, 12 rebuilds

I understand that on on sbt multi-module build, you chose to record one buildinfo per artifact/module, where I chose to record 1 aggregated buildinfo, each strategy having pros and cons

but I'd really like to avoid maintaining 3*4 buildspecs: can you try to at least strip to 3 buildspecs, even if each buildspec records 4 buildinfo (they are generated, then not really that important)?
And in the end, generate 1 buildinfo.compare result (comparison against Central Repository content) for these 4 buildinfo

I don't know if in the future consolidating the 3 buildspecs into a unique one with multiple minor scala versions would make sense: let's go step by step

raboof:
> 12 buildspecs, 12 rebuilds

I agree that seems wasteful :)

> can you try to at least strip to 3 buildspecs, even if each buildspec records 4 buildinfo

Building multiple buildinfo's from one buildspec seems fine to me. I wonder what the path structure convention should look like, then, though.

AFAICS, the ultimate goal is to answer the question: "I downloaded this artifact, can I trust it?". For example, when I'm building a project that depends on `com.scalapenos:stamina-json_2.12:1.5.0`, I want to find the `.buildinfo`s to confirm it is untampered with. An easy way to achieve that would be to place the `.buildinfo` at a path that mirrors the path of the original artifact - so in this example `content/com/scalapenos/stamina-core_2.12/0.1.5/stamina-core_2.12-0.1.5.buildinfo`.

Right now `buildinfo.sh` copies `${buildinfo}` to `../..`. This works for single-buildinfo builds, but what if we want to generate buildinfo's at multiple paths? Perhaps we should allow specifying in the `buildspec` at what path the `buildinfo` should end up?

hboutemy:
> AFAICS, the ultimate goal is to answer the question: "I downloaded this artifact, can I trust it?".
yes, this is ultimate goal. But we're not yet there.

In this step, I wanted to make sure there was reproducible content available, that anybody was able to rebuild, whatever the build tool (I was not able to rebuild your artifacts, for example, that was frustrating and made clear that nobody would reproduce...).

The key learning here is the requirement for **buildspec**: we need to make them available and usable by rebuilders to check that reproducibility is effective or not, whoever did the reference release in his personal environment (I had some surprises on some of my artifacts).

The directory structure is based on the buildspec granularity, reusing groupId and "root" artifactId ("root" meaning human interpretation). I did not put version as a directory but as a file. I stored output buildinfo in the same directory as buildspec as Maven produces 1 buildinfo per buildspec (even for multi-module build), but perhaps this should be changed given sbt produces many buildinfo per buildspec...

The output currently is the README table report to show to humans how many projects provide reproducible artifacts: it does not really go at artifact level, because it would add too much data, but when you see a result with tens or hundreds of output, I let you imagine

I'm confident that other output formats can be automated from this step, that will permit the target usage we want. But we have time before sufficient reproducible content is available.

raboof:
> In this step, I wanted to make sure there was reproducible content available, that anybody was able to rebuild

I guess it's a matter of perspective, starting with the end in mind vs starting from where we are - similar to top-down vs bottom-up, they'll eventually meet ;)

> I stored output buildinfo in the same directory

OK, I can PR that for stamina

> I finally added buildinfo.compare because comparison against reference artifacts is the real objective here

I just noticed the `-Dreference.repo` and `-Dreference.compare.save` options you're passing to maven - I didn't realize before that's something provided by the maven plugin. I'd be happy to add something similar to the sbt plugin (should be fairly simple since there's already the `reproducibleBuildsCheckMavenCentral` target). I'd like to iterate on the file format a bit, though.

> I'm confident that other output formats can be automated from this step, that will permit the target usage we want.

OK. Should we eventually add that as an additional step, exploding the buildinfo's in `content` (organized by groupId and "root artifact id") to a tree mirroring the maven artifact locations, organized by groupId and artifactId? Perhaps as a first step simply in a separate tree in this repo?

hboutemy:
> similar to top-down vs bottom-up, they'll eventually meet ;)
+1000 :)

> I'd like to iterate on the file format a bit, though.
yes, this is when using that we discover a few improvements, like this `-Dreference.compare.save` that I added because of the current reproducible-central needs

> Should we eventually add that as an additional step, exploding the buildinfo's in content (organized by groupId and "root artifact id") to a tree mirroring the maven artifact locations, organized by groupId and artifactId? Perhaps as a first step simply in a separate tree in this repo?
yes, good idea: the target gavs will have to point to the buildspec that produced them
This will give us the 2 key use cases:
1. as a dependency user, I want to know if the artifact is reproducible
2. as a rebuilder, I want to rebuild a release and check that I can reproduce
This is the meet in the middle we were looking for :)

hboutemy:
superseded by #12

PR #12, opened 2020-05-03
Use 1 buildspec to build all stamina artifacts

raboof:
This does not yet work with `update-reproducibility-summary.sh`.

Not sure if we should have a `.compare` file per buildinfo or per buildspec?

To supersede #11


hboutemy:
looks really nice: but you forgot to commit rebuild.sh
and one thing that is surprising is that the choice of the 3 minor scala version sis not part of the buildspec: is this list is really never meant to be updated?

the `.compare` is ideally per buildspec, to display in the report in README.md

raboof:
> you forgot to commit rebuild.sh

I didn't make changes to rebuild.sh ;)

> the choice of the 3 minor scala version is not part of the buildspec

It is part of the sbt configuration inside the project, so it's nice we don't have to specify it in the buildspec anymore

> the `.compare` is ideally per buildspec, to display in the report in README.md

Makes sense, I updated `update-reproducibility-summary.sh` to derive the `.compare` filename from the `.buildspec` and added a manually-created .compare for stamina.

hboutemy:
> It is part of the sbt configuration inside the project, so it's nice we don't have to specify it in the buildspec anymore

oh yes, great!

> Makes sense, I updated `update-reproducibility-summary.sh` to derive the `.compare` filename from the `.buildspec`
no, that won't work given how multi-module buildinfo are not associated to "root" artifactId but more to the last one in the build: this is too cumbersome to write the buildspec

I'll merge as much as possible and we'll see what's the next step

hboutemy:
I think that the algorithm for `update-reproducibility-summary.sh` should be:
- if only 1 buildinfo file, add `.compare` to its filename (= like previously)
- if multiple buildinfo files for the buildspec, derive compare filename from buildspec

this should give sensible results for both strategies, isn't it?

hboutemy:
done in cc648e63f241ddaa998daba949c8ef90cc75950f
notice the `.compare` content, that lists 24 reproducible files

raboof:
> I think that the algorithm for `update-reproducibility-summary.sh` should be:

Seems fine to me. If we add the convention that if a build only produces 1 buildinfo file, the name of the buildspec should match the name of the buildinfo, then we can always derive the compare filename from the buildspec. But this works fine too :+1:

hboutemy:
> If we add the convention that if a build only produces 1 buildinfo file, the name of the buildspec should match the name of the buildinfo

theoretically, you're right. But if I did not implement it like this, that's not just because I chose to be more complex than the obvious wish.
I don't know how multi-module builds work in sbt, but with Maven, the root pom build happens first before every other modules: then at root execution time, we don't have any output content yet to fingerprint. Fingerprinting of everything can only happen on the last module. That's why my aggregated buildinfo does not happen on the module we would theoretically expect

PR #32, opened 2020-08-06
Initial CI setup

bhamail:
This is still a WIP, but by following the steps in the `.circleci/circleci-readme.md`, you can run a CI build locally.

After installing the circleci client, you can run a local build via:

```
circleci config process .circleci/config.yml > .circleci/local-config.yml && circleci local execute -c .circleci/local-config.yml --job 'build-one'
```

Docker volumes are not mounting in a CircleCI build. see: https://support.circleci.com/hc/en-us/articles/360007324514-How-can-I-use-Docker-volume-mounting-on-CircleCI-
Found this: https://circleci.com/docs/2.0/building-docker-images/#mounting-folders
Changed the `rebuild.sh` script to user a "data" container and "docker cp" to move app files to the data container before the build, and then after the build to copy output out of the data container.


hboutemy:
having a look at this again.
Building everything has no real interest as of today: too much workload to re-do what was already done
Building a job parametrized with a buildspec target can be useful, to have a rebuild different from local rebuild. The question is then to show if the CI rebuild gives the same result as the local rebuild. And I have one key question: who can run the job?

one thing that could help me is to have a daily run of `update-reproducibility-summary.sh` that checks in updated `README.md`, because this is currently something I do manually and is cumbersome

bhamail:
Makes sense.
Will look into `update-reproducibility-summary.sh`, but I'm guessing such a "nightly" update will need access to the completed builds? (or can it just rely on stuff that will already be banked?) Guess I should "use the source" and answer the question myself.  ;)

hboutemy:
@bhamail should we close that PR now that we have the CI rebuilding README.md?

bhamail:
Fixed by #63

PR #34, opened 2020-09-17
buildspec for maven-site-plugin 3.9.1.

bhamail:
It looks like the m-site-p [GH workflow](https://github.com/apache/maven-site-plugin/blob/maven-site-plugin-3.9.1/.github/workflows/maven.yml) uses JDK8, so I tried bumping the .buildspec to 8 and it seems to have solved the differences.

fixes issue #17

Ready to create PR to add reproducible-build-badge here: https://github.com/bhamail/maven-site-plugin/tree/badge-reproducible-build
I figured I should wait until this buildspec is accepted.

### DIFF
--- a/content/org/apache/maven/plugins/maven-site-plugin/maven-site-plugin-3.9.1.buildinfo
+++ b/content/org/apache/maven/plugins/maven-site-plugin/maven-site-plugin-3.9.1.buildinfo
@@ -1,0 +1,37 @@
+# https://reproducible-builds.org/docs/jvm/
+buildinfo.version=1.0-SNAPSHOT
+
+name=Apache Maven Site Plugin
+group-id=org.apache.maven.plugins
+artifact-id=maven-site-plugin
+version=3.9.1
+
+# source information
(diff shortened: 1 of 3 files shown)


hboutemy:
merged, thank you Dan

FYI, when I rebuilt locally, I saw that by buildinfo file was not exactly the same as yours: `java.version=1.8.0_242` in mine, when in yours `java.version=1.8.0_265`
everything important is the same = the fingerprints
but this just proves that rebuilding locally may give a different buildinfo on this detailed java version line: I don't know yet if we should do something about it or not in this Reproducible Central project...

bhamail:
> merged, thank you Dan
> 
> FYI, when I rebuilt locally, I saw that by buildinfo file was not exactly the same as yours: `java.version=1.8.0_242` in mine, when in yours `java.version=1.8.0_265`
> everything important is the same = the fingerprints
> but this just proves that rebuilding locally may give a different buildinfo on this detailed java version line: I don't know yet if we should do something about it or not in this Reproducible Central project...

Interesting. Not sure I have a strong opinion on how best to handle the build jvm version diff.

Issue #24, opened 2020-07-13
create buildspec for maven-war-plugin 3.3.1


hboutemy:
should be as reproducible as 3.3.0...

PR #35, opened 2020-09-17
buildspec for maven-war-plugin 3.3.1.

bhamail:
Dropped jdk from 8 to 7 and that appears to fix the diffs.

Fixes issue #24

Looks like the badge is already added to the m-war-p project.


hboutemy:
> Dropped jdk from 8 to 7 and that appears to fix the diffs

FYI, in fact, on each release, the release manager who publishes the official release binaries to Central does his own choice on which JDK he is using: then on each release, rebuilding requires to find which JDK was used to reuse same major version

PR #52, opened 2021-01-06
buildspec for org.simplify4u.plugins:pgpverify-maven-plugin

slawekjaranowski:


hboutemy:
merged buildspec
then rebuilt and published result in 2cb807f614dc10ecdbd9d853e53546dec4e22fab

thank you

PR #54, opened 2021-01-12
improve colors of project badge

Krinkle:
Use a darker blue for the left side, matching the dark blue of the [reproducible-builds.org](https://reproducible-builds.org/) logo.

Use a different green for the right side, matching the standard green used for other "success" badges from <https://shields.io/>, such as for continuous integration, dependency updates, and license checks.

| Before | After
|--|--
| [Reproducible Builds](https://reproducible-builds.org/) | [Reproducible Builds](https://reproducible-builds.org/)

I've adopted this in <https://github.com/qunitjs/qunit#readme> as example.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,10 +55,10 @@ between reference file from Central Repository and effective rebuild file, then 
 
 <details><summary><b>Add Reproducible Builds Badge to a Project With Reproducible Releases</b></summary>
 
-If a project has listed here at least one release with proven reproducibility success, it can add a badge like ![Reproducible Builds](https://img.shields.io/badge/Reproducible_Builds-ok-green?labelColor=blue) pointing to its entries here:
+If a project has listed here at least one release with proven reproducibility success, it can add a badge like ![Reproducible Builds](https://img.shields.io/badge/Reproducible_Builds-ok-success?labelColor=1e5b96) pointing to its entries here:
 
 ```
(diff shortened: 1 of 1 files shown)


hboutemy:
nice, thank you for your help

I see your project is JavaScript, it would be nice to share practices between ecosystems.
I have 2 questions:
- is there a location where Reproducible Builds for JavaScript practices are described?
- how do you check that your project release is reproducible?

Krinkle:
@hboutemy I'm not aware of a generalised approach or solution that is or can be shared by the JavaScript ecosystem at large (or even specific to those published as NPM package).

For QUnit specifically, my base level involves the distributed JS artefact file to have identical contents between what we publish on NPM, what we host on our CDN, and what one can independently generate using the following steps:

1. clean checkout of the Git repository at the release tag.
2. environment has a supported version of Node.js and npm (Node 10+ and npm 6+ at time of writing).
3. Set `SOURCE_DATE_EPOCH="$(git log -s --format=%at -1)"`, as per the [spec](https://reproducible-builds.org/docs/source-date-epoch/) and our [release doc](https://github.com/qunitjs/qunit/blob/2.14.0/RELEASE.md).
4. Run `npm ci && npm run build`

… after which the artifact file is (or shoudl be) byte-for-byte identical to what was published.

Issue #42, opened 2020-10-11
create new buildspecs for projects that define Reproducible Builds timestamp for Maven build


hboutemy:
here are some POMs in Central Repository:
com/samboom/samboom/1/samboom-1.pom
com/github/ldapchai/ldapchai/0.7.6/ldapchai-0.7.6.pom
com/io7m/primogenitor/com.io7m.primogenitor/6.0.0/com.io7m.primogenitor-6.0.0.pom
io/dwpbank/movewp3/microservice-security-autoconfiguration/0.6.0/microservice-security-autoconfiguration-0.6.0.pom
org/apache/jackrabbit/filevault-package-maven-plugin/1.1.4/filevault-package-maven-plugin-1.1.4.pom
org/apache/sling/org.apache.sling.api/2.23.0/org.apache.sling.api-2.23.0.pom
org/apache/struts/struts2-parent/2.5.25/struts2-parent-2.5.25.pom
org/apache/accumulo/accumulo-project/1.10.0/accumulo-project-1.10.0.pom
org/apache/maven/plugins/maven-ear-plugin/3.1.0/maven-ear-plugin-3.1.0.pom
org/apache/maven/plugins/maven-resources-plugin/3.2.0/maven-resources-plugin-3.2.0.pom
org/apache/maven/shared/maven-filtering/3.2.0/maven-filtering-3.2.0.pom
org/eclipse/transformer/org.eclipse.transformer.parent/0.2.0/org.eclipse.transformer.parent-0.2.0.pom
org/osgi/org.osgi.test.parent/0.10.0/org.osgi.test.parent-0.10.0.pom

each require a PR for corresponding buildspec

hboutemy:
this manula way of discovering what to rebuild is now replaced by daily run of `checkOutputTimestamp()` https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/update_artifact_badge_data.java#L58
scheduled in https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/.github/workflows/update-reports.yml

PR #58, opened 2021-04-27
#42 Maven-scripting-plugin v3.0.0

dialaya:
Add maven-scripting-plugin, maven-jxr


hboutemy:
great work, thank you

PR #59, opened 2021-04-29
typo

bhamail:
example command missing `.sh` suffix on script name


hboutemy:
good catch, thank you

PR #63, opened 2021-06-25
Work In Progress: initial nightly CI job

bhamail:
Here's a first stab at a nightly job to update the summary page, as described in: https://github.com/jvm-repo-rebuild/reproducible-central/pull/32.

Current status is it doesn't work yet. ;)

The CircleCI "shellcheck" tool suggested a number of tweaks (see the diff of `update-reproducibility-summary.sh`) mostly along the lines of replacing backticks and using double quotes. I probably broke something, but at least with the tweaks, I get the script to run (in a local CI build) with this error:

```shell
$ circleci config process .circleci/config.yml > .circleci/local-config.yml && circleci local execute -c .circleci/local-config.yml --job 'job-nightly'

...
====>> ./update-reproducibility-summary.sh

  #!/bin/sh -eo pipefail
./update-reproducibility-summary.sh

*** running script: ./update-reproducibility-summary.sh
ls: content/com/scalapenos/stamina/stamina*_*-0.1.5.buildinfo: No such file or directory
./update-reproducibility-summary.sh: line 99: content/com/scalapenos/stamina/stamina*_*-0.1.5.buildinfo.compare: No such file or directory
ls: content/com/scalapenos/stamina/stamina*_*-0.1.6.buildinfo: No such file or directory
./update-reproducibility-summary.sh: line 99: content/com/scalapenos/stamina/stamina*_*-0.1.6.buildinfo.compare: No such file or directory
Error: Exited with code 1
Step failed
Error: runner failed (exited with 101)
Task failed
Error: task failed
``` 

After I get the shell script to succeed, I can dig into how to update the summary page in github.

Since I was chasing my tail for a while debugging the build, I added a little message to the script indicating the build managed to at least execute the script (which will hopefully keep me from getting dizzy in the future).

If you want to see the "warnings" the produced by `shellcheck` just uncomment this bit of the `config.yml`:

```
#      - shellcheck/install
#      - shellcheck/check:
#          pattern: update-reproducibility-summary.sh
```

FWIW, I'm fine with rolling back all the shellcheck changes, but figured they might make it easier to debug.


hboutemy:
stamina is the only Scala project, with some special cases: I fear that your fixes are just showing how the script worked previously by chance...

hboutemy:
I searched what is the root cause: SBT project buildspec require to define multiple values for "buildinfo" value, see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/scalapenos/stamina/stamina-0.1.6.buildspec

such a case works well for rebuild.sh, that uses the value to copy from build workspace to result directory: usually the result is only 1 file, but copying multiple files is fine
but it fails in update-reproducibility-summary.sh, which uses the value to define only one result file to load from

I see 2 ways of fixing:
- improve update-reproducibility-summary.sh to detect such exceptional case
- update SBT project buildspec and rebuild.sh to split "buildinfo" into "buildinfo" (with only 1 resulting file) and "buildinfo_extra"

I don't really see how to implement first option, the transparent one
implementing the second option is quite simple...

bhamail:
"buildinfo_extra" doesn't sound so bad. (I also don't know how to implement option 1)

update-reproducibility-summary.sh
```diff
@@ -8,22 +10,22 @@ countGa=0
...
-  . $buildspec
+  . "$buildspec"
   new_anchor="${groupId}:${artifactId}"
-  [[ -z "${issue}" ]] && [[ ! -z "${diffoscope}" ]] && issue="$(dirname ${buildspec})/$(basename ${diffoscope})"
+  [[ -z "${issue}" ]] && [[ -n "${diffoscope}" ]] && issue="$(dirname "${buildspec}")/$(basename "${diffoscope}")"
 
-  buildinfo="`dirname ${buildspec}`/`basename ${buildinfo}`"
-  if [ `ls ${buildinfo} | wc -l` -le 1 ]; then
+  buildinfo="$(dirname "${buildspec}")/$(basename "${buildinfo}")"
+  if [ $(ls "${buildinfo}" | wc -l) -le 1 ]; then
```
hboutemy, line 25: ok, I found the real root cause: please remove quotes
`ls "${buildinfo}"` need to be replaced with `ls ${buildinfo}`

bhamail: cool. I made that change in b527dc2, and now I'm seeing a different error (with less output). will see if I can find the cause of that.

bhamail: I think I figured out it is just the last line of the script that is causing a non-zero return value from `update-reproducibility-summary.sh`:

```shell
grep "unexpected issue" README.md
```

Am I correct in thinking we actually "want" that command to "error" out, meaning everything is happy if that command does not find the text "unexpected issue" in README.md?

bhamail: Assuming I'm not totally missing the point (a major assumption), I changed the `grep` command in e7a6495

hboutemy: great job, thanks

hboutemy:
the updated script now works like a charm: it's nice that the error code is more consistent now: when running locally, I never really investigated why my prompt said there was a failure, when I got the perfect result in README.md :)

now, last steps are:
1. if no issue, commit README.md
2. activate scheduled run

I don't know what is about content of the commit and what is about me configuring my GH project: please advise

bhamail:
I was thinking next steps would be to have the "nightly CI"job automatically:
1. ~~run `update-reproducibility-summary.sh`~~ done
2. check for changes made to `README.md` (after running `update-reproducibility-summary.sh`)
3. commit changes made to `README.md`
4. push `README.md` changes to github

(Might even be able to skip step 2 - if nothing changed, "git commit,push" might be a no-op)

I should be able to use a github deploy key (with `write` access) to do the above in CircleCI. Can safely configure that via protected env vars, etc. (I'll pester you for that key if needed).

I plan to add these steps to the CI config next (if no objections/corrections).

To summarize, my thinking was the "nightly CI" build will update README.md once per day automagically.

PR #64, opened 2021-08-05
document requirement for gsed on mac

bhamail:
still testing to make sure the documented solution solves the problem, but looks correct so far.

Also starting to feel like  the "Prerequisites" section to move up to being a top level bullet, as I think it applies to all subsequent bullets.

Piling on: I saw this error when running the `build_diffoscope.sh` command: 
```
sed: ip/maven-shared-utils-3.3.4.diffoscope: No such file or directory
```
I'm guessing it is because the default `sed` is not gnu. I tried to steal what you did in the other script.

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,17 @@ equivalent to the packaging instructions that are maintained by every Linux dist
     * `dos2unix` - DOS to MAC/UNIX text file format converter. 
       
       Can be installed via [homebrew](https://brew.sh) on MAC via: `brew install dos2unix`.
+ 
+    * `gsed` - on mac, to provide gnu sed command. 
+      
+      Can be installed via [homebrew](https://brew.sh) on MAC via: `brew install gsed`.
+ 
+      If you are missing `gsed` on a MAC, you may see an error like the one below:
+          
+          ...
+          dos2unix: converting file target/maven-shared-utils-3.3.4.buildinfo to Unix format...
+          dos2unix: converting file target/maven-shared-utils-3.3.4.buildinfo.compare to Unix format...
+          ./rebuild.sh: line 143: gsed: command not found
    </details>
 
 You can rebuild a project release by running:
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -20,6 +20,17 @@ equivalent to the packaging instructions that are maintained by every Linux dist
     * `dos2unix` - DOS to MAC/UNIX text file format converter. 
       
       Can be installed via [homebrew](https://brew.sh) on MAC via: `brew install dos2unix`.
+ 
+    * `gsed` - on mac, to provide gnu sed command. 
+      
+      Can be installed via [homebrew](https://brew.sh) on MAC via: `brew install gsed`.
+ 
+      If you are missing `gsed` on a MAC, you may see an error like the one below:
```
hboutemy, line 28: I think that this part can be removed, given the next comment...

build_diffoscope.sh
```diff
@@ -7,15 +7,21 @@ diffoscope_file=$(dirname ${compare})/$(basename ${compare} .buildinfo.compare).
 
 echo -e "saving build diffoscope file to \033[1m${diffoscope_file}\033[0m"
 
-grep '# diffoscope ' ${compare} | sed -e 's/# diffoscope //' | ( while read -r line
+set sed="sed"
+if [ "$(uname -s)" ==  "Darwin" ]
+then
```
hboutemy, line 12: we can reuse the existence check pattern to bring nice error message:
```
command -v gsed >/dev/null 2>&1 || { echo "require GNU sed: brew install gsed  Aborting."; exit 1; }
```

hboutemy (changes requested):

README.md
bhamail: nice. removed 2034885

build_diffoscope.sh
bhamail: makes sense to me. done 2034885

hboutemy (approved):

PR #65, opened 2021-08-05
maven-shared-utils 3.3.4 diffs help

bhamail:
I created this PR against my prior branch to minimize the noise.

I'm not sure what to try next to clean up the remaining diffs. Any suggestions?

I have intentionally not moved the files into the final location because this is still work-in-progress. Hence: NOTE TO SELF: DO NOT MERGE AS IS. :)


hboutemy:
interesting case: your wrote the best buildspec possible, selecting as expected JDK 8 on Linux, but the output bytecode of your run has a few unexpected differences with the reference
Personally, I just launched the rebuild with provided buildspec, and the output does not have the difference: my host is Linux, and the Docker image used provided Oracle JDK 1.8.0_275

can you check which precise JDK version was used in your rebuild?
I suppose that javac evolved a little bit in a patch version

The reproducibility environment summary "any JDK 8" may be not perfect: until now, it worked quite well given the 100s of successful rebuilds, but it would be useful to document the rare failures to see how we should enhance the env specification (if it happens more and more over time)
FYI, I had 1 equivalent case that I stored as "not reprocible", but I should take time to try with different JDK 11 minor versions: there are not so many releases, it should be feasible https://github.com/jvm-repo-rebuild/reproducible-central/blob/maven-shared-utils_3.3.4/content/io/fabric8/docker-maven-plugin/docker-maven-plugin-0.36.1.diffoscope

hboutemy:
just tested from my Mac: I get the same sha256 than you, different from reference build from Central...
Host OS is Mac, JDK version is Oracle 1.8.0_262
I will try to see if I can update the Docker image used to match my Linux host, to see if the subtle difference is due to Docker image/JDK bugfix version or if it is due to Mac host...

hboutemy:
I updated my Docker image to get Oracle JDK 1.8.0_282
same issue on Mac host than with 1.8.0_262

uh, javac run inside a container does not get the same result based on the host OS: completely unexpected!!!

hboutemy:
content merged, with a note on the Mac issue: ideally, we should check if this Mac javac output issue is recurring or purely random

if recurring, we'll need at least to document on the rebuild requirements that Mac has this issue (and in fact make clear that the rebuilds were done on Linux hosts, but for example never tested either with Windows hosts...)

we should probably close this issue as merged and open another issue if we find a strategy around this Mac output issue

bhamail:
works for me. happy to have this PR closed out if that makes sense to you.

PR #71, opened 2022-02-03
Add pmd 6.42.0

### LINKED ISSUE #70: fix recent PMD rebuilds
PMD until 6.32 is perfectly reproducible, with buildspec provided to prove the rebuilds

next releases could not be reproduced, because there is a new Kotlin dependency that breaks the build:
```
$ ./add-new-release.sh content/net/sourceforge/pmd/pmd-6.32.0.buildspec 6.42.0
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  24.345 s
[INFO] Finished at: 2022-01-30T08:44:15Z
[INFO] ------------------------------------------------------------------------
[ERROR] Missing:
[ERROR] ----------
[ERROR] 1) org.jetbrains.kotlinx:kotlinx-html-jvm:jar:0.7.2
[ERROR] 
[ERROR]   Try downloading the file manually from the project website.
[ERROR] 
[ERROR]   Then, install it using the command: 
[ERROR]       mvn install:install-file -DgroupId=org.jetbrains.kotlinx -DartifactId=kotlinx-html-jvm -Dversion=0.7.2 -Dpackaging=jar -Dfile=/path/to/file
[ERROR] 
[ERROR]   Alternatively, if you host your own repository you can deploy the file there: 
[ERROR]       mvn deploy:deploy-file -DgroupId=org.jetbrains.kotlinx -DartifactId=kotlinx-html-jvm -Dversion=0.7.2 -Dpackaging=jar -Dfile=/path/to/file -Durl=[url] -DrepositoryId=[id]
[ERROR] 
[ERROR]   Path to dependency: 
[ERROR]         1) org.jetbrains.dokka:dokka-base:jar:1.4.30
[ERROR]         2) org.jetbrains.kotlinx:kotlinx-html-jvm:jar:0.7.2
...
```

we need to define where these artifacts come from and update `settings.xml`, ...

adangel:
* Fixes #70 

We still need jcenter (https://github.com/pmd/pmd/blob/master/pom.xml#L931-L943). Adding this to settings.xml made the build work.


content/net/sourceforge/pmd/pmd-6.42.0.buildspec
```diff
@@ -0,0 +1,16 @@
...
+version=6.42.0
+
+gitRepo=https://github.com/pmd/${artifactId}.git
+gitTag=pmd_releases/${version}
+
+tool=mvn
+jdk=11
+newline=lf
+
+command="mvn -Ppmd-release clean package -DskipTests -Dmaven.javadoc.skip -Dgpg.skip"
```
adangel, line 13: we actually use maven wrapper - so would this be more appropriate in terms of reproducibility?

```suggestion
command="./mvnw -Ppmd-release clean package -DskipTests -Dmaven.javadoc.skip -Dgpg.skip"
```

hboutemy: good idea, yes, to easily match your env in a reproducible way

hboutemy:
merged, thank you

mvn vs mvnw will really make a difference if some time in the future the exact version of Maven used makes a difference: in general, it does not, that's why it's not a parameter that I added to .buildspec

PR #83, opened 2022-07-22
Fix a syntax error in sample .buildspec

Pfiver:
This would throw here: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/rebuild.sh#L22

```
$ sourceRmFiles=DEPENDENCIES LICENSE NOTICE
bash: LICENSE: command not found
```


hboutemy:
nice finding, thank you

Issue #90, opened 2022-10-08
Combination of Maven 3.8.6 and JDK 17 not supported


Stephan202:
When specifying `tool=mvn-3.8.6` and `jdk=17`, `rebuild.sh` will try to pull image `maven:3.8.6-openjdk-17-slim` as a result of [this logic](https://github.com/jvm-repo-rebuild/reproducible-central/blob/36870ee8f893926c8bf47e761d0c5fca0c3f7de1/rebuild.sh#L119-L121), but that image [does not exist](https://hub.docker.com/_/maven/tags?page=1&name=3.8.6-openjdk-17-slim):
> docker: Error response from daemon: manifest for maven:3.8.6-openjdk-17-slim not found: manifest unknown: manifest unknown.

For newer Maven releases, should the script use one of the [other distributions](https://hub.docker.com/_/maven/tags?page=1&name=3.8.6) instead?

hboutemy:
what is hard is that supported tags on Maven DockerHub images is evolving over time and JDK releases https://hub.docker.com/_/maven , then the logic to define image name is hard to maintain with all its parameters: Maven version and JDK version are the main ones, but selected image has to guess a JDK vendor and OS distro. See https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/rebuild.sh#L104

perhaps we could support some parameters to let buildspec explicitely choose its Docker image

any idea?

notice: each time I change the default JDK vendor and OS distro to have a new Maven version + JDK version work, I'm not sure it works with all previous Maven + JDK combination that worked until now
I confess it's a magic in the rebuild script that would probably require some unit tests: I suppose that we could write a script to extract from all existing buildspecs which Maven + JDK versions are used, then test that all combinations work...

Stephan202:
> perhaps we could support some parameters to let buildspec explicitely choose its Docker image

I think this would make sense (and can't think of another solution that doesn't risk breaking other builds, either). :+1: 

An additional advantage of allowing users to explicitly specify the tag, is that they could even specify the image digest, such that the build environment itself is also frozen/fully reproducible. (But the downside of pinning the image digest is that it is likely that over time this repo would reference _a lot_ of distinct images, which presumably impacts the storage and runtime requirements of the automated rebuild cron job.)

hboutemy:
help welcome

> such that the build environment itself is also frozen/fully reproducible

honestly, experience shows that we don't need very precise rebuild environment

[...]

PR #93, opened 2022-10-22
Use more robust Maven image selection process

Stephan202:
This is a proposal based on @hboutemy's suggestion [here](https://github.com/jvm-repo-rebuild/reproducible-central/issues/90#issuecomment-1287831003).

### DIFF
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -102,13 +102,34 @@ mvnBuildDocker() {
   esac
 
   # select Docker image to match required JDK version: https://hub.docker.com/_/maven
-  mvnImage=$(
-    printf "%s\n" maven:{${mvnVersion},3}-{jdk,openjdk,eclipse-temurin}-${jdk}{-alpine,-slim,} \
-      | while read image; do
-          docker manifest inspect ${image} > /dev/null 2>&1 && echo ${image}
-        done \
-      | head -n 1
-  )
+  case ${jdk} in
+    6 | 7)
+      mvnImage=maven:3.6.1-jdk-${jdk}-alpine
+      crlfDocker="yes"
+      ;;
+    8)
+      mvnImage=maven:${mvnVersion}-jdk-${jdk}-slim
+      ;;
+    9)
+      mvnImage=maven:3-jdk-${jdk}-slim
+      ;;
+    14)
+      mvnImage=maven:${mvnVersion}-jdk-${jdk}
+      ;;
+    11 | 15 | 16)
+      mvnImage=maven:${mvnVersion}-openjdk-${jdk}-slim
+      ;;
+    *)
+      mvnImage=maven:${mvnVersion}-eclipse-temurin-${jdk}-alpine
+  esac
+  if ! docker pull -q ${mvnImage} > /dev/null 2>&1; then
+    for image in maven:{${mvnVersion},3}-eclipse-temurin-${jdk}-alpine; do
+      if docker pull -q ${image} > /dev/null 2>&1; then
+        mvnImage=${image}
+        break
+      fi
+    done
+  fi
 
   echo "Rebuilding using Docker image ${mvnImage}"
   local docker_command="docker run -it --rm --name rebuild-central -v $PWD:/var/maven/app -v $base:/var/maven/.m2 -v $base/.sbt:/var/maven/.sbt -v $base/.npm:/.npm -v $base/.bnd:/.bnd -u $(id -u ${USER}):$(id -g ${USER}) -e MAVEN_CONFIG=/var/maven/.m2 -w /var/maven/app"
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -102,26 +102,13 @@ mvnBuildDocker() {
   esac
 
   # select Docker image to match required JDK version: https://hub.docker.com/_/maven
-  case ${jdk} in
-    6 | 7)
-      mvnImage=maven:3.6.1-jdk-${jdk}-alpine
-      crlfDocker="yes"
-      ;;
-    8)
-      mvnImage=maven:${mvnVersion}-jdk-${jdk}-slim
-      ;;
-    9)
-      mvnImage=maven:3-jdk-${jdk}-slim
-      ;;
-    14)
-      mvnImage=maven:${mvnVersion}-jdk-${jdk}
-      ;;
-    11 | 15 | 16)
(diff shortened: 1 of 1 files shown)


hboutemy:
good idea

I tried to evaluate the impact: i added https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/update-statistics.sh script that generates https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/doc/buildspec-stats.txt

I have an issue with current full rewrite: it takes ages to run and has a huge impact

I imagine that keeping the existing algorithm would be ok, with adding at the end a check that the image exists, and if it does not exist, using the new algorithm with less options (because we don't need old values for old JDKs)

we would have the best of all worlds, isn't it?

Stephan202:
Works for me! I added a commit with a new proposal. Note that I swapped out  `docker manifest inspect` for `docker pull -q`, because the former doesn't seem faster for existing images, while the latter performs an operation required by the subsequent `docker run`.

rebuild.sh
```diff
@@ -122,6 +122,14 @@ mvnBuildDocker() {
     *)
       mvnImage=maven:${mvnVersion}-eclipse-temurin-${jdk}-alpine
   esac
+  if ! docker pull -q ${mvnImage} > /dev/null 2>&1; then
+    for image in maven:{${mvnVersion},3}-eclipse-temurin-${jdk}-alpine; do
```
Stephan202, line 126: For recent JDKs and Maven versions these two variants _seem_ to suffice.

rebuild.sh
```diff
@@ -122,6 +122,14 @@ mvnBuildDocker() {
...
+      if docker pull -q ${image} > /dev/null 2>&1; then
+        mvnImage=${image}
+        break
+      fi
```
Stephan202, line 130: I realized that the previous revision could inspect more images than necessary. Are simple for-loop avoids that.

rebuild.sh
```diff
@@ -122,6 +122,14 @@ mvnBuildDocker() {
     *)
       mvnImage=maven:${mvnVersion}-eclipse-temurin-${jdk}-alpine
   esac
+  if ! docker pull -q ${mvnImage} > /dev/null 2>&1; then
```
hboutemy, line 125: @Stephan202 I tested this (with bin/update-statistics.sh) and found that `docker pull -q` fails when the image is already available

Stephan202: Interesting; locally I can't reproduce this:
```sh
$ docker --version
Docker version 20.10.20, build 9fdeb9c
$ docker pull -q maven:3.8.5-eclipse-temurin-17-alpine 2>&1 2>/dev/null; echo $?
docker.io/library/maven:3.8.5-eclipse-temurin-17-alpine
0
$ docker pull -q maven:3.8.5-eclipse-temurin-17-alpine 2>&1 2>/dev/null; echo $?
docker.io/library/maven:3.8.5-eclipse-temurin-17-alpine
0
$ docker pull -q maven:this-one-does-not-exist 2>&1 2>/dev/null; echo $?
1
```

Perhaps it depends on the version of Docker used (which one do you have installed locally?).

I can have a closer look at `bin/update-statistics.sh` later; right now I'm commuting; don't want to pull new images while using mobile internet :)

Stephan202: > I can have a closer look at `bin/update-statistics.sh` later.

Didn't happen yesterday, and today is full of meeting and events; I intend to circle back to this PR later this week. Sorry for the delay.

hboutemy: no worries, we all have a day job, I have the same type of delays that you patiently waited for

hboutemy: I just tested (and improved a little bit the script): you're right, `docker pull -q` works, I probably was hit by my rate limit or anything stupid and transient

now, with the `bin/update-statistics.sh` script, we can easily evaluate the impact of any change in image tag selection and see if it breaks any buildspec tool x jdk selection

I let you review and update the statistics script while changing `rebuild.sh` so we can merge with safety

Stephan202: Finally found a moment to get back to this :)

I saw that `bin/update-statistics.sh` uses `docker inspect` rather than `docker manifest inspect`. So this means there are three options. I wrote the following script to compare them:
```sh
#!/usr/bin/env bash

# The image against which to test.
image=maven:3.6.3-openjdk-17-slim

print_header() {
  local header="${1}"
  echo -e "\n\033[0;34m${header}\033[0m"
}

print_result() {
  test $? == 0 && echo -e '\033[0;32mSUCCESS\033[0m' || echo -e '\033[0;31mFAILURE\033[0m'
}

# Drop the image if it exists.
docker image rm -f "${image}"

print_header 'Running `docker inspect` while the image is locally absent.'
time docker inspect "${image}" > /dev/null
print_result

print_header 'Running `docker manifest inspect` while the image is locally absent.'
time docker manifest inspect "${image}" > /dev/null
print_result

print_header 'Running `docker pull -q` while the image is locally absent.'
time docker pull -q "${image}" > /dev/null
print_result

print_header 'Running `docker inspect` while the image is locally present.'
time docker inspect "${image}" > /dev/null
print_result

print_header 'Running `docker manifest inspect` while the image is locally present.'
time docker manifest inspect "${image}" > /dev/null
print_result

print_header 'Running `docker pull -q` while the image is locally present.'
time docker pull -q "${image}" > /dev/null
print_result
```

Based on the results, it seems to me that:
1. `docker inspect` is by far the fastest, **but** only says whether the given image exists locally. It seems this command cannot be used in `rebuild.sh`, unless there's some kind of "pre-population" of images. (Which might be worth the effort of implementing; see below.)
2. `docker manifest inspect` takes ~3 seconds if the image doesn't exist, and ~5 if it does, _irrespective of how often the command is executed_.
3. `docker pull -q` takes ~2 seconds if the image doesn't exist, and ~1.5 if is does, _unless the image doesn't exist locally yet, in which case it of course takes longer_ (but that's a cost that needs to be paid regardless).

Based on this I see two options:
1. Stick with `docker pull -q` in `rebuild.sh`.
2. Introduce a `docker pull -q`-based "bootstrap phase" and use `docker inspect` in `rebuild.sh`.

(The timings presented here _may_ of course be specific to my setup; interested to hear whether you can reproduce these results.)

hboutemy:
yes, great: I merged and reworked the algorithm into 0b5d8be9f78fc84b60c11804102f68a0dc89f632

thanks a lot

PR #97, opened 2022-11-13
Podman engine support

nhumblot:
Hi :wave: 

Thank you for this repository and your work on reproducible builds. I'm proposing this change to enable a podman support as a container engine instead of docker. It might help users to work on this project by providing an alternative for a container engine.

This change has been made as a non-blocking change. Existing commands are expected to still use docker as a default, without any change.

The flexibility has been added with an allow list in mind to prevent users specifying an unplanned and untested alternative. This is performed in ` rebuild.sh` through the `container_engine` variable and the `containerEngine()` function, introduced by this PR.

I worked on this change to be minimal. If you would prefer a more robust change with a named parameter, do not hesitate to let me know. Your comments are welcomed, thank you for your help :slightly_smiling_face:

### DIFF
--- a/rebuild.sh
+++ b/rebuild.sh
@@ -86,15 +86,6 @@ fi
 
 echo -e "\033[1m$(pwd)\033[0m"
 
-containerEngine() {
-  if [[ ${container_engine} == "podman" ]]
-  then
-    podman "$1"
-  else
-    docker "$1"
-  fi
-}
-
 mvnBuildDocker() {
   local mvnCommand mvnImage crlfDocker mvnVersion
   mvnCommand="$1"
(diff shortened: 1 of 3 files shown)


rebuild.sh
```diff
@@ -86,6 +86,15 @@ fi
 
 echo -e "\033[1m$(pwd)\033[0m"
 
+containerEngine() {
```
hboutemy, line 89: given `${container_engine}` is used in other location, this function that is used in only one location is a duplicate: I feel it's better to drop it in favor of `${container_engine}` for every caller, isn't it?

nhumblot: Yes, definitely! Thank you for your comment. I removed this method which is of no use. I still have an error while executing the command, so I'm keeping this PR as a draft until I find a way to make it work :)

PR #102, opened 2023-01-07
add an entry on cyclonedx-maven-plugin to the Buildspec 'Tips and Tricks' section

marcwrobel:


hboutemy:
very good addition, thanks you

PR #103, opened 2023-02-12
[DRAFT] buildspec for Yauaa

nielsbasjes:
Hi,

I need advise from you.

This project ( https://github.com/nielsbasjes/yauaa ) needs JDK 8, 11 AND 17 installed with the correct toolchains config to build it.

How do I achieve this?

Also: When a project is part of this. What effort is needed on a new release? Has this been automated?

PR #104, opened 2023-02-14
feat: Support for multi JDK builds

nielsbasjes:
This is a draft for supporting Toolchains with all LTS JDK versions: 8, 11 and 17
Included a basic test using this project https://github.com/nielsbasjes/ToolChainsInCiBuilds/

NOTE: This project has NOT been released to maven central so it WILL fail because it has nothing to compare with and this project was never intended to be reproducible.

You should be able to do this and it should do `a lot` with all 3 different JDK versions.

    ./rebuild.sh wip/toolchains-1.0.buildspec

This is NOT yet perfect.

The problem I keep bumping my head against is that it keeps seeing '?' as being the user.home in various situations.

If I remove the ` -t /var/maven/.m2/toolchains.xml` parameter and make the command `mvn clean -X` I see this:

> [DEBUG] Reading global settings from /usr/share/maven/conf/settings.xml
> [DEBUG] Reading user settings from /var/maven/.m2/settings.xml
> [DEBUG] Reading global toolchains from /usr/share/maven/conf/toolchains.xml
> [DEBUG] Reading user toolchains from ?/.m2/toolchains.xml
> [DEBUG] Using local repository at /var/maven/.m2/repository
> [DEBUG] Using manager EnhancedLocalRepositoryManager with priority 10.0 for /var/maven/.m2/repository

with the focus on the directory of the user toolchains: `?/.m2/toolchains.xml`
Note that I do have `-Duser.home=/var/maven` on the commandline.

This is Apache 2 licensed code. So have fun with it.


nielsbasjes:
Just to be clear: This is activated in the buildspec by setting

   jdk=lts

You will get an image with all LTS versions and a working toolchains config.

nielsbasjes:
I think the `-Duser.home=/var/maven` should not be a commandline parameters.
Mainly because sometimes a maven build runs a script which is then unaware of this setting.

nielsbasjes:
By setting `jdk=lts` in the spec file this `rebuild.sh` script will run maven under Java 17 and provide all 3 current LTS JDKs (8, 11 and 17) via toolchains.
This works on my machine when trying to build the spec file I posted here: #103 

Would like to hear your feedback on this idea/approach.

nielsbasjes:
FYI: @hboutemy

rebuild.sh
```diff
@@ -87,18 +194,21 @@ fi
...
 mvnBuildDocker() {
-  local mvnCommand mvnImage crlfDocker mvnVersion
+  local mvnCommand mvnImage crlfDocker mvnVersion mvnDefaultJDKVersion
   mvnCommand="$1"
   crlfDocker="no"
 
   mvnVersion="3.6.3"
+  mvnDefaultJDKVersion=${jdk}
   case ${jdk} in
-    18)
```
hboutemy, line 96: no, please don't change existing defaults: they are used by existing buildspecs

rebuild.sh
```diff
@@ -116,41 +226,97 @@ mvnBuildDocker() {
     14)
       mvnImage=maven:${mvnVersion}-jdk-${jdk}
       ;;
-    11 | 15 | 16 | 17)
```
hboutemy, line 119: no, please don't change existing defaults: they are used by existing buildspecs

rebuild.sh
```diff
@@ -172,7 +338,7 @@ mvnBuildLocal() {
 rebuildToolMvn() {
   # the effective rebuild command, adding artifact:buildinfo goal to compare with central content
   #local mvn_rebuild="${command} -V -e artifact:compare -Dbuildinfo.reproducible"
-  local mvn_rebuild="${command} -V -e org.apache.maven.plugins:maven-artifact-plugin:3.3.0:compare -Dbuildinfo.reproducible"
```
hboutemy, line 175: I'd prefer have this change in a separate PR

hboutemy (changes requested):
thanks Niels, very nice now that I found time to deeply review

I'd love this PR to be split in 2:
1. preparation, with the introduction of your runcommand/info/fatal/pass/warn functions
2. then the Maven Toolchain LTS JDKs-specific image feature

on the second aspect, I think that we could have something more extensible by not hijacking `jdk` value in `.buildspec` with a magic `lts` case (that does not represent the Maven Toolchain aspect), but instead add an additional `image` parameter which by default is `auto` that triggers the current image selection based on jdk+buildtool, but when set with a different value in `.buildspec` would prepare a specific image from a collection stored in `bin/images` subdirectories

WDYT?

rebuild.sh
nielsbasjes: I recently found that Java 18 no longer is available while fixing the Apache Avro build.
https://github.com/apache/avro/commit/12adf0bc57cb1ddcb58a4f6aa41695fb7512140a

So I think that none of the buildspecs should have anything other than an LTS JDK version.

rebuild.sh
nielsbasjes: I agree: Defaults must remain the same.

rebuild.sh
nielsbasjes: Ok

wip/toolchains-1.0.buildspec
```diff
@@ -0,0 +1,27 @@
...
+display=${groupId}:${artifactId}
+
+# Source code
+gitRepo=https://github.com/nielsbasjes/ToolChainsInCiBuilds.git
+gitTag=main
+
+# Rebuild environment prerequisites
+tool=mvn-3.8.7
+# All LTS JDK versions and JDK-17 is the default (used to starts Maven with)
+jdk=lts
```
nielsbasjes, line 15: I'm not happy with this part yet.
Indicating `lts` makes it unclear which exact versions are needed. I expect in a few years that the term `lts` means something different (i.e. no more jdk 8).
So the exact set of needed JDK versions must be made explicit.

Also because of the timezone and umask issues I think these also must be made explicit. 
If unspecified I want to set 
- the timezone to `Australia/Eucla`: a real inhabited timezone with (as far as I can tell) the smallest population (~200 people).
- the umask to 0055 : Which makes absolutely no sense in a real situation.

So either a project clearly indicates the needed values for reproduction, or the project has all plugins configured correctly and it builds in a reproducible way. 

In general: If it is relevant for the reproducibility it must be specified in the buildspec config.

So I'm now thinking about something like:
```
# The jdk under which maven must be run
jdk=17
# The jdk versions available via toolchains
toolchain=8|11|17
# The default timezone that is to be used by the JDK (i.e. java -Duser.timezone=..)
timezone=Europe/Amsterdam
# The umask that must be set during the build
umask=0002
```

rebuild.sh
hboutemy: > So I think that none of the buildspecs should have anything other than an LTS JDK version

we have no choice: if the release manager did a bad choice that does not last over time, we can't change his choice

wip/toolchains-1.0.buildspec
hboutemy: that's why in the second step, I propose to name your specific image, and there will be other images

on umask, it's another topic: it will be another step

we need to go step by step

rebuild.sh
nielsbasjes: Yes you are right.

Yet consider: how long after a "bad choice" was made does this project need to be able to support proving the reproducibility of it? I say there should be an plausible end to it.
https://endoflife.date/java

hboutemy: experience shows that some projects choose random JDK over releases, which is not an issue for the content (given `-release x` guarantees that there is no issue in the bytecode) but for long term reproducibility, makes environment creation harder.
Notice: harder, not impossible...

hboutemy: and we have the statistics of such fragile choices https://github.com/jvm-repo-rebuild/reproducible-central#tool-x-jdk-statistics

nielsbasjes:
Important:
- I have tested several projects with different JDK versions and they still work (on my machine)
- I have not tested on Mac OS / Windows / Cygwin and such. Only on Linux.
- I noticed that the .buildinfo file created here is not reproducible. I tried apache-rat 0.15 and found that the ordering of the files is random. So you get diffs like this:
```
-outputs.5.2.filename=apache-rat-0.15-bin.zip
-outputs.5.2.length=2572347
-outputs.5.2.checksums.sha512=1e5dda3a378be437c4d349e3a383ce23916401c48f0dd65aea703aadfd255863e949e08aed709cb64024fd01ba859e8b2083d1ce8b7fc11f96dfea2a551ed531
+outputs.5.2.filename=apache-rat-0.15-src.zip
+outputs.5.2.length=525203
+outputs.5.2.checksums.sha512=effad21b9329db53174b4886e72119e234a0b656ebf477ea2b6c1f02bc65b9d06b23155cf7a15663651c17171e9134474dc66db0e4829e597d65bb6ef9a92202
```

Notable changes:
- [CHANGE] Only downloads the specified image. If a build asks for a specific maven version for which there is no image it will simply fail (and not try a different version).
- You can prepend the command with `DOCKER` (similar to `SHELL`) yet now you get a command prompt IN the docker environment.
- You can now specify `toolchains`, `locale`, `umask`, `timezone` and extra `MAVEN_OPTS`.
So now the nifi 1.20.0 can be built automatically using something like this:
```
groupId=org.apache.nifi
artifactId=nifi
display=${groupId}:${artifactId}
version=1.20.0

gitRepo=https://github.com/apache/${artifactId}.git
gitTag=rel/${artifactId}-${version}

tool=mvn
jdk=8
newline=lf
locale=en_US
MAVEN_OPTS='-Duser.country=US -Duser.language=en'

command="mvn -Papache-release,include-grpc,include-ranger clean package -DskipTests -Dmaven.javadoc.skip -Dgpg.skip"
buildinfo=target/${artifactId}-${version}.buildinfo

diffoscope=${artifactId}-${version}.diffoscope
issue=https://issues.apache.org/jira/browse/NIFI-10892
```

nielsbasjes:
@hboutemy This still needs some cleanup but it already works quite nicely.
You can try the wip/yauaa-7.15.0.buildspec (reproducible and published to maven central) and wip/toolchains-1.0.buildspec (not published to central) which shows the new tooling at work.

nielsbasjes:
Note that this also adds the user to ALL images and then does the mount.
This way it all works a lot better and you can open a shell into any of those images by prefixing the command with `DOCKER `. Also the `-Duser.home=/var/maven` is no longer needed.

nielsbasjes:
A bonus feature that works if you have a few jdks in toolchains and you have the docker shell open:
```bash
 nbasjes@[Reproducible Builder <JDK 1.8.0_362>]:~/app {main} ]
$ java -version
openjdk version "1.8.0_362"
OpenJDK Runtime Environment (build 1.8.0_362-8u362-ga-0ubuntu1~22.04-b09)
OpenJDK 64-Bit Server VM (build 25.362-b09, mixed mode)
 nbasjes@[Reproducible Builder <JDK 1.8.0_362>]:~/app {main} ]
$ SwitchJDK 11
Setting JDK to version 11
 nbasjes@[Reproducible Builder <JDK 11.0.18>]:~/app {main} ]
$ java --version
openjdk 11.0.18 2023-01-17
OpenJDK Runtime Environment (build 11.0.18+10-post-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 11.0.18+10-post-Ubuntu-0ubuntu122.04, mixed mode)
 nbasjes@[Reproducible Builder <JDK 11.0.18>]:~/app {main} ]
$ SwitchJDK 17
Setting JDK to version 17
 nbasjes@[Reproducible Builder <JDK 17.0.6>]:~/app {main} ]
$ java --version
openjdk 17.0.6 2023-01-17
OpenJDK Runtime Environment (build 17.0.6+10-Ubuntu-0ubuntu122.04)
OpenJDK 64-Bit Server VM (build 17.0.6+10-Ubuntu-0ubuntu122.04, mixed mode, sharing)
 nbasjes@[Reproducible Builder <JDK 17.0.6>]:~/app {main} ]
$ SwitchJDK 18
[FAIL] JDK 18 is not available
[INFO] Available are:
java-1.11.0-openjdk-amd64      1111       /usr/lib/jvm/java-1.11.0-openjdk-amd64
java-1.17.0-openjdk-amd64      1711       /usr/lib/jvm/java-1.17.0-openjdk-amd64
java-1.8.0-openjdk-amd64       1081       /usr/lib/jvm/java-1.8.0-openjdk-amd64
 nbasjes@[Reproducible Builder <JDK 17.0.6>]:~/app {main} ]
$ 
```

hboutemy:
PR rebased to local branch https://github.com/jvm-repo-rebuild/reproducible-central/tree/ToolChains
I'll test it now for a few days before merging to master: looks really promising

nielsbasjes:
Good to hear.
Looking forward to the results.
If you want me to test something or fix something please let me know.

hboutemy:
merged in PR #106  from branch with additional fixes found when testing on a Mac
thanks a lot @nielsbasjes , this is an awesome improvement

PR #107, opened 2023-03-24
Add conventional-commits-version-policy

nielsbasjes:
and a tiny fix in the .gitignore


.gitignore
```diff
@@ -5,6 +5,7 @@ settings-docker.xml
 .sbt/
 .npm/
 .bnd/
+.docker/
```
hboutemy, line 8: this one is not necessary: you should delete your local directory, current scripts don't store files here

nielsbasjes: Ok, I removed this change.

PR #108, opened 2023-04-06
add jreleaser 1.6.0

aalmiray:


aalmiray:
Converting to draft as JReleaser `1.6.0` has yet to be released.

aalmiray:
Additionally, the `jreleaser` module creates a JAR and Zip/Tar distributions. I'd prefer if the `jreleaser-<version>.zip` artifact found inside `jreleaser/build/distributions` to be the chosen artifact to be checked, instead of the JAR.

aalmiray:
@hboutemy do you have any ideas on how to check the distribution Zip instead of the JAR?

hboutemy:
I don't really understand the question: everything that is published to Maven Central is supposed to be checked. I suppose it's `jreleaser*` from https://repo.maven.apache.org/maven2/org/jreleaser/

see script: https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/includes/rebuildToolGradle.sh
it's the content that should be in the `repository/` directory after the build, that will be compared to content from fetched from Maven Central to `central/` directory

aalmiray:
The JReleaser build produces additional artifacts such as Zip and a Tar file that _are_ _not_ published to Maven Central.

hboutemy:
not published to central, then not detected as part of the output of the release that is included in buildinfo, and not checked by the current script that requires first to download reference content (from Maven Central...)

if you see a repeatable strategy for extending beyond Maven Central binaries, don't hesitate to share and we'll eventually try to implement in this "Reproducible Central" effort

aalmiray:
@hboutemy Gotcha. The alternative would be to download a matching binary (in this case the Zip archive) from a given location. Most common location would be a release asset from a Github/Gitlab release, in which case the spec would have to support at least one additional section where such artifact (local path, remote URL) could be defined.

For the time being let's go with what Reproducible Central offers, which is very close to what I need except that:
- the launcher script won't be checked as it's not published with JARs but with the Zip.
- there's the potential of additional local JARs & POMs that won't match with Maven Central. JReleaser publishes 2 Gradle plugins which I think will be push to maven local but won't be published to Maven Central, thus the checker might find local archives without matches as it performs a broad search of all artifacts matching a given `groupId`, does it not?

hboutemy:
> Most common location would be a release asset from a Github/Gitlab release, in which case the spec would have to support at least one additional section where such artifact (local path, remote URL) could be defined

you're right, we could probably add some config to buildspec to define a location to download reference from, and a path to local build output to define the scope to compare

this will be a future enhancement: thinking at it, it's exactly the initial objective of JReleaser that is not covered = CLI that are published to "OS" package managers (Brew, etc...)
JReleaser could even have its own separate "Reproducible JReleaser" Git repo (eventually under current `jvm-repro-rebuild` GH org) to target this scope

aalmiray:
I believe you've successfully nerd-snipped me into building `reproducible-releases` 😱

hboutemy:
you mean this one https://github.com/jvm-repo-rebuild/reproducible-releases that you've just been invited to?

aalmiray:
@hboutemy this is ready for review now

hboutemy:
just tried to rebuild but it failed:

```
❯ ./rebuild.sh content/org/jreleaser/jreleaser/jreleaser-1.6.0.buildspec

FAILURE: Build failed with an exception.

* Where:
Settings file '/var/gradle/app/settings.gradle' line: 47

* What went wrong:
A problem occurred evaluating settings 'app'.
> Failed to apply plugin class 'org.kordamp.gradle.plugin.inline.InlinePlugin'.
   > /var/gradle/app/%3F/.gradle/caches/jars-9/b5dfc3f665468a7103b563ef33c3a54a/shadow-7.1.2.jar
```

any idea what's going wrong on my local build? I suppose it worked for you

aalmiray:
That's odd. Java 11 and the local gradle wrapper should be enough to build the project. I'll run another check.

hboutemy:
same issue with this update: did you try `./rebuild.sh content/org/jreleaser/jreleaser/jreleaser-1.6.0.buildspec`?

Reading the error message, the `%3F` directory name in `/var/gradle/app/%3F/.gradle/caches/jars-9/b5dfc3f665468a7103b563ef33c3a54a/shadow-7.1.2.jar` looks strange

aalmiray:
I can replicate the problem

```
Caused by: java.nio.file.NoSuchFileException: /var/gradle/app/%3F/.gradle/caches/jars-9/b5dfc3f665468a7103b563ef33c3a54a/shadow-7.1.2.jar
        at java.base/sun.nio.fs.UnixException.translateToIOException(UnixException.java:92)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:111)
        at java.base/sun.nio.fs.UnixException.rethrowAsIOException(UnixException.java:116)
        at java.base/sun.nio.fs.UnixFileAttributeViews$Basic.readAttributes(UnixFileAttributeViews.java:55)
        at java.base/sun.nio.fs.UnixFileSystemProvider.readAttributes(UnixFileSystemProvider.java:149)
        at java.base/sun.nio.fs.LinuxFileSystemProvider.readAttributes(LinuxFileSystemProvider.java:99)
        at java.base/java.nio.file.Files.readAttributes(Files.java:1764)
        at java.base/java.util.zip.ZipFile$Source.get(ZipFile.java:1259)
        at java.base/java.util.zip.ZipFile$CleanableResource.<init>(ZipFile.java:733)
        at java.base/java.util.zip.ZipFile$CleanableResource.get(ZipFile.java:850)
        at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:248)
        at java.base/java.util.zip.ZipFile.<init>(ZipFile.java:177)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:350)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:321)
        at java.base/java.util.jar.JarFile.<init>(JarFile.java:260)
        at org.kordamp.gradle.plugin.inline.InlinePlugin.extractMetadata(InlinePlugin.groovy:219)
```

It seems to me that `InlinePlugin` performs a path check that breaks when Gradle Home is set to a different value than the default

https://github.com/kordamp/kordamp-gradle-plugins/blob/70883eac6bd9608f6494be01120b553ae7c87757/plugins/inline-gradle-plugin/src/main/groovy/org/kordamp/gradle/plugin/inline/InlinePlugin.groovy#L219

In any case, fixing or disabling this plugin requires changing code and thus JReleaser `v1.6.0` can't be checked for reproducibility just yet 😓

aalmiray:
Well, I spoke too soon. I managed to find a way to make the build work without any code changes. Yet, there are files published to MC which are not reproducible, mainly cyclonedx SBOMs in JSON and XML formats.

aalmiray:
Filed https://github.com/CycloneDX/cyclonedx-gradle-plugin/issues/292

@hboutemy Do you know if https://github.com/CycloneDX/cyclonedx-maven-plugin generates reproducible SBOMs?

hboutemy:
this will happen in next release, I just implemented it https://github.com/CycloneDX/cyclonedx-maven-plugin/pull/353

aalmiray:
Great! Well, this leaves is with JReleaser itself for now. I've configured the gradle command line to skip publishing 2 gradle plugins as they are not available from Maven Central. Cyclonedx SBOMs are not reproducible and they are also published to Maven Central. 

Question: should we skip checking SBOMs for now as we know they are not reproducible? Or do we wait until we can ensure SBOMs are reproducible and we try again?

hboutemy:
we should skip checking SBOMs = what I do with Maven (by adding `-Dcyclonedx.skip` to build command)

aalmiray:
I'll have to check if the gradle plugin offers a similar capability.

aalmiray:
Unfortunately there's no option available in the Gradle plugin to disable SBOM generation using a system/project property. I'm afraid that's not a common idiom for Gradle plugins (but it should be).

I can add a workaround for a future build yet, that does not solve the problem for `1.6.0`. The workaround for this version would be to run a script that deletes SBOMs from the local repository before diff checks are performed. 

WDYT?

aalmiray:
Managed to configure JReleaser `1.7.0` to skip publishing SBOMs by default.
But for `1.6.0` the problem may not be solved. I thought that setting a compound command on the spec would work, such as

```
command="./gradlew --no-daemon -PreproducibleBuild=true -Dorg.kordamp.gradle.inline.enabled=false \
  -S publishToMavenLocal -x :jdks-gradle-plugin:publishToMavenLocal -x :jreleaser-gradle-plugin:publishToMavenLocal \
  && find /var/gradle/.m2/repository/org/jreleaser -name "*.cyclonedx.*" -exec rm {} \;"
```

but the calling scripts says this

```
  local gradle_docker_params="-Duser.home=/var/gradle"
  runcommand ${docker_command} ${jdkImage} ${command} ${gradle_docker_params}
```

which means the actual command looks like

```
./gradlew --no-daemon -PreproducibleBuild=true -Dorg.kordamp.gradle.inline.enabled=false \
-S publishToMavenLocal -x :jdks-gradle-plugin:publishToMavenLocal -x :jreleaser-gradle-plugin:publishToMavenLocal \
&& find /var/gradle/.m2/repository/org/jreleaser -name "*.cyclonedx.*" -exec rm {} \; -Duser.home=/var/gradle
```

Don't want to change the common commands for this to work. I think it would be best to wait for `1.7.0` which I have tested locally with OK results.

hboutemy:
yes, working to have better future releases will avoid loosing time on existing releases: past is past...

PR #109, opened 2023-04-22
add logparser 5.9.0

nielsbasjes:
I call the project "logparser" https://github.com/nielsbasjes/logparser but none of the deployed artifacts actually have this name.

Please verify if I have done this correctly this way.


content/nl/basjes/parse/logparser/logparser-5.9.0.buildspec
```diff
@@ -0,0 +1,31 @@
+# Central Repository coordinates for the Reference release (for multi-module, pick an arbitrary module)
+groupId=nl.basjes.parse
+artifactId=logparser
```
hboutemy, line 3: this artifactId does not exist in Maven Centrral: https://repo.maven.apache.org/maven2/nl/basjes/parse/
this will disable maven-metadata.xml check that detect if newer releases are available
you should rename to [parser-parent](https://repo.maven.apache.org/maven2/nl/basjes/parse/parser-parent/)

nielsbasjes: I have changed it to the artifact that I have documented people should use.
```
<dependency>
    <groupId>nl.basjes.parse.httpdlog</groupId>
    <artifactId>httpdlog-parser</artifactId>
    <version>5.9.0</version>
</dependency>
```

and the `maven-metadata.xml` now contains the correct information (on my machine).

PR #110, opened 2023-04-24
Update yauaa-7.17.0.buildspec

nielsbasjes:
Add link to the issue of the root cause of this version not being reproducible.


hboutemy:
notice: does it mean that by building with Maven 3.9.2 (not released yet), we could reproduce the build?

nielsbasjes:
Yes, I expect that to be the case.

PR #114, opened 2023-05-25
Add jstachio 1.0.0

agentgt:
Please add jstachio 1.0.0


agentgt:
@hboutemy I wasn't sure if I should move the files from the wip directory to content?

It appears to build fine with a minor hack I had to do with time zone (ant doesn't UTC on updating archives so it had my America/New_York tz).

wip/jstachio-1.0.0.buildspec
```diff
@@ -0,0 +1,21 @@
...
+version=1.0.0
+
+display=${groupId}:${artifactId}
+
+gitRepo=https://github.com/jstachio/${artifactId}.git
+gitTag=v${version}
+
+tool=mvn-3.9.2
+jdk=17
+newline=lf
```
hboutemy, line 12: need to add `umask=022`
my own laptop has 002 default umask, and this makes a difference in result

wip/jstachio-1.0.0.buildspec
```diff
@@ -0,0 +1,21 @@
...
+
+# Unfortunately one of the Ant calls to update jstachio-apt jar did not normalize the timestamp on UTC
+# This will be fixed in the next version
+command="bin/rebuild.sh -Duser.timezone=America/New_York"
+
+# Location of the buildinfo file generated during rebuild to record output fingerprints
+buildinfo=api/jstachio/target/${artifactId}-${version}.buildinfo
```
hboutemy, line 19: define `buildinfo=target/jstachio-maven-parent-${version}.buildinfo` to get the aggregated full build output for 31 files, not only 1 random module with 3 files

hboutemy (changes requested):
requiring exact 17.0.7 JDK release makes it hard to rebuild (I could not on my laptop, it uses 17.0.6)
can this be avoided?

hboutemy:
nice work: yes, keeping it in `wip/` while we're working on it is perfect

agentgt:
I’ll have the changes tomorrow. I will switch it to 1.0.1 of jstachio and use all your recommendations.

My rebuild.sh happen to be an accidental similar name.

It executes a Java file as if it were a script to update the POM. This is because I find the release plugin to not meet my needs.

hboutemy:
sorry for the long delay, I completely forgot to review
merged, and new releases added like a charm (with my usual `bin/add-new-release.sh` script)

well done, thanks

PR #115, opened 2023-06-08
Buildspec for pl.com.labaj:auto-record:1.1.0

pawellabaj:


hboutemy:
cherry-picked the update, thank you

FYI, I had to add a umask 022 configuration, that is your default but not mine: this shows I'll need to improve the rebuild script to better manage this...

PR #123, opened 2023-08-14
Add io.github.shanqiang-sq;jstream;1.0.31

Vel1khan:


hboutemy:
last learning from this case: it's useful to also store `maven-metadata.xml`, that will be used to detect new releases

Vel1khan:
@hboutemy  Thanks for the feedback. Will be including maven-metadata.xml in the future. For the tag, do you mean replacing ```gitTag=jstream-root-1.0.31``` with ```gitTag=jstream-root-{$version}```?

PR #130, opened 2023-10-27
auto-record groupId changed

pawellabaj:
Created new buildspec after groupId changed

I'm facing problems with building in Podman - see pawellabaj/auto-record#133


hboutemy:
please don't remove history: history remains, even if in a now defunct groupId
I just did the groupId update for release 3.0.0 in 8b7718e74bbc7734d1bcde50b02043b0327caee2

I also published release 3.1.0

I proposed you https://github.com/pawellabaj/auto-record/pull/134 to configure skip deploy in a more classical way

on your issue with podman, are you sure the right Maven version is used?

pawellabaj:
Thanks for proposed changes in pom.

As to build issues, I tried default maven and the latest - "mvn-3.9.5" - and I got the same issues

pawellabaj:
@hboutemy , how to ignore some artifacts in `.buildspec`? 
I don't plan to release test or doc artifacts

hboutemy:
with PR merged, it should be automatic
without PR merged, it requires a new maven-artifact-plugin release because of https://issues.apache.org/jira/browse/MARTIFACT-52

hboutemy:
as you can see, 3.1.1 is ok https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/pl/com/labaj/autorecord/auto-record/README.md
at least i can rebuild it: is it ok to you?

pawellabaj:
It's ok

Thank you @hboutemy for help

PR #134, opened 2023-12-18
Add `commons-logging:commons-logging`

ppkarwasz:
The `commons-logging` artifacts version 1.3.0 are fully reproducible except:

 * the SPDX descriptor.

Due to moditect/moditect#222 the timezone is one of the variables to take into consideration. In this case it is `America/New_York`.

**Remark**: all the artifacts that use `commons-parent` version 65 (the first one with Moditect version 1.1.0) should be reproducible too.


hboutemy:
thanks a lot for the buildspec and the moditect analysis

PR #138, opened 2024-01-19
fix url of instructions in buildspec template

sebastian-toepfer:


hboutemy:
> is -Dcyclonedx.skip still needed?
no, RB for this plugin has been implemented (in 2.7.9 IIRC)

PR #149, opened 2024-04-14
add ServerLib-2.3.6

NotMyFault:
I'm not entirely sure if the files generated are complete, due to https://github.com/jvm-repo-rebuild/reproducible-central/issues/150.

Please let me know if something is missing.

### DIFF
--- a/content/dev/notmyfault/serverlib/ServerLib/ServerLib-2.3.6.buildcompare
+++ b/content/dev/notmyfault/serverlib/ServerLib/ServerLib-2.3.6.buildcompare
@@ -1,0 +1,5 @@
+version=2.3.6
+ok=4
+ko=0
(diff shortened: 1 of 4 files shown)


hboutemy:
thank you: I'll rebuild to check

hboutemy:
I rebuilt: I get the same result, only the order of files is different, that aspect should probably be improved in https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/includes/compareOutput.sh

NotMyFault:
> I rebuilt: I get the same result, only the order of files is different, that aspect should probably be improved in [`master`/bin/includes/compareOutput.sh](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/includes/compareOutput.sh?rgh-link-date=2024-04-14T21%3A11%3A35Z)

Instead of `du -b`, `du -B 1` would work (for me at least), which specifies the block size, which -b [aliases](https://man7.org/linux/man-pages/man1/du.1.html) for.

Issue #150, opened 2024-04-14
Comparing results on MacOS yields error


NotMyFault:
While working on https://github.com/jvm-repo-rebuild/reproducible-central/pull/149, I noticed the following issue within the comparison script:

```
[INFO] computing ServerLib-2.3.6.buildinfo from content in repository/ directory, downloading equivalent from Maven Central to central/, and comparing: results to ServerLib-2.3.6.buildcompare
du: invalid option -- bo central/dev/notmyfault/serverlib/ServerLib/2.3.6/ServerLib-2.3.6-sources.jar                              
usage: du [-Aclnx] [-H | -L | -P] [-g | -h | -k | -m] [-a | -s | -d depth] [-B blocksize] [-I mask] [-t threshold] [file ...]
du: invalid option -- bo central/dev/notmyfault/serverlib/ServerLib/2.3.6/ServerLib-2.3.6.jar                              
usage: du [-Aclnx] [-H | -L | -P] [-g | -h | -k | -m] [-a | -s | -d depth] [-B blocksize] [-I mask] [-t threshold] [file ...]
du: invalid option -- bo central/dev/notmyfault/serverlib/ServerLib/2.3.6/ServerLib-2.3.6.module                              
usage: du [-Aclnx] [-H | -L | -P] [-g | -h | -k | -m] [-a | -s | -d depth] [-B blocksize] [-I mask] [-t threshold] [file ...]
du: invalid option -- bo central/dev/notmyfault/serverlib/ServerLib/2.3.6/ServerLib-2.3.6.pom                              
usage: du [-Aclnx] [-H | -L | -P] [-g | -h | -k | -m] [-a | -s | -d depth] [-B blocksize] [-I mask] [-t threshold] [file ...]

[INFO] rebuild from wip/serverlib-2.3.6.buildspec
```
which might have mangled the files generated in my PR.

hboutemy:
notice that surprisingly, even if there is an error message, the value in file is ok

PR #152, opened 2024-04-15
Update 'du -b' for MacOS users

NotMyFault:
Picks up https://github.com/jvm-repo-rebuild/reproducible-central/pull/149#issuecomment-2054190512. The change proposed works on my MacOS machine and on ubuntu 22.04.

Fixes https://github.com/jvm-repo-rebuild/reproducible-central/issues/150


bin/includes/compareOutput.sh
```diff
@@ -58,7 +58,7 @@ compareOutput() {
       wget -q https://repo.maven.apache.org/maven2/$f -O central/$f
 
       echo "outputs.${n}.${i}.filename=$(basename $f)" >> ${buildinfo}
-      echo "outputs.${n}.${i}.length=$(du -b ${outputDir}/$f | cut -f 1)" >> ${buildinfo}
+      echo "outputs.${n}.${i}.length=$(du -B 1 ${outputDir}/$f | cut -f 1)" >> ${buildinfo}
```
hboutemy, line 61: on Linux, it round to inode size
requires `--apparent-size` too: is this option ok on MacOS?

NotMyFault: Doesn't look so, according to the manual:

```
➜  Desktop man du

     -A      Display the apparent size instead of the disk usage.  This can be helpful when operating on compressed volumes or sparse files.

	-B blocksize
             Calculate block counts in blocksize byte blocks.  This is different from the -h, -k, -m, --si and -g options or setting BLOCKSIZE and gives an estimate of how much space the examined file hierarchy would require on a filesystem with the given blocksize.
             Unless in -A mode, blocksize is rounded up to the next multiple of 512.

     -H      Symbolic links on the command line are followed, symbolic links in file hierarchies are not followed.

     -I mask
             Ignore files and directories matching the specified mask.

     -L      Symbolic links on the command line and in file hierarchies are followed.

     -P      No symbolic links are followed.  This is the default.

     -a      Display an entry for each file in a file hierarchy.

     -c      Display a grand total.

     -d depth
             Display an entry for all files and directories depth directories deep.

     -g      Display block counts in 1073741824-byte (1 GiB) blocks.

     -h      “Human-readable” output.  Use unit suffixes: Byte, Kilobyte, Megabyte, Gigabyte, Terabyte and Petabyte based on powers of 1024.

     -k      Display block counts in 1024-byte (1 kiB) blocks.

     -l      If a file has multiple hard links, count its size multiple times.  The default behavior of du is to count files with multiple hard links only once.  When the -l option is specified, the hard link checks are disabled, and these files are counted (and
             displayed) as many times as they are found.

     -m      Display block counts in 1048576-byte (1 MiB) blocks.

     -n      Ignore files and directories with user “nodump” flag (UF_NODUMP) set.

     -r      Generate messages about directories that cannot be read, files that cannot be opened, and so on.  This is the default case.  This option exists solely for conformance with X/Open Portability Guide Issue 4 (“XPG4”).

     -s      Display an entry for each specified file.  (Equivalent to -d 0)

     --si    “Human-readable” output.  Use unit suffixes: Byte, Kilobyte, Megabyte, Gigabyte, Terabyte and Petabyte based on powers of 1000.

     -t threshold
             Display only entries for which size exceeds threshold.  If threshold is negative, display only entries for which size is less than the absolute value of threshold.

     -x      File system mount points are not traversed.

     The du utility counts the storage used by symbolic links and not the files they reference unless the -H or -L option is specified.  If either the -H or -L option is specified, storage used by any symbolic links which are followed is not counted (or displayed).  The
     -H, -L and -P options override each other and the command's actions are determined by the last one specified.

     Files having multiple hard links are counted (and displayed) a single time per du execution.  Directories having multiple hard links (typically Time Machine backups) are counted a single time per du execution.  The -h, -k, -m and --si options all override each
     other; the last one specified determines the block counts used.

ENVIRONMENT
     BLOCKSIZE  If the environment variable BLOCKSIZE is set, and the -h, -k, -m or --si options are not specified, the block counts will be displayed in units of that block size.  If BLOCKSIZE is not set, and the -h, -k, -m or --si options are not specified, the block
                counts will be displayed in 512-byte blocks.
```

hboutemy: doing portable shell between Linux and Mac is definitively more complex than it looks...

hboutemy: yes, we'll probably have to detect OS and change the command used to get file size

for example like done at the beginning of https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/detect-new-releases.sh with the command to get date

hboutemy:
great, thank you

PR #160, opened 2024-04-21
Add podman support for Gradle builds

jskov:
Hi there!

I run a Fedora system with SELinux enabled, and use Podman as a container engine.

So I have changed the rebuild and gradle scripts to allow using podman.

While I can reproduce the build for:

```console
REBUILD_ENGINE=podman REBUILD_VOLUME_FLAGS=":Z,rw" ./rebuild.sh content/org/mockito/mockito-core/mockito-5.8.0.buildspec
```

I am unsure if the changes might have broken the docker (gradle) support?

In particular, there may be some trouble around the mounting of .gradle.
Not sure why it was mounted as it was before? The folder did not exist.
Instead I mount the location of the folder created above (similar to .m2).

### DIFF
--- a/bin/includes/rebuildToolGradle.sh
+++ b/bin/includes/rebuildToolGradle.sh
@@ -26,9 +26,9 @@ rebuildToolGradle() {
 #      jdkImage="openjdk:17-slim"
 #      ;;
 #  esac
-  jdkImage="gradle:8-jdk${jdk}"
+  jdkImage="docker.io/library/gradle:8-jdk${jdk}"
 
-  info "Rebuilding using Docker image ${jdkImage}"
+  info "Rebuilding using image ${jdkImage}"
 
   local OUTPUTDIR=repository
   [ -d ${OUTPUTDIR} ] && \rm -rf ${OUTPUTDIR}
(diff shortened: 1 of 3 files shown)


hboutemy:
I'm starting with https://www.ericsbinaryworld.com/2019/11/02/can-docker-and-podman-both-run-on-the-same-machine/ ...

jskov:
No need to apologize.

In fact, it would probably be reasonably for you to ask me to at least make the same changes for Maven.
And I will be happy to (have better time now).

Then you could just verify that things still work with docker?

hboutemy:
if you can, that would be awesome, thanks

rebuild.sh
```diff
@@ -87,6 +87,25 @@ umask $umask
 export MVN_UMASK=${umask}
 fetchSource
 
+DEFAULT_engine="docker"
+DEFAULT_engineopts_docker=""
+DEFAULT_engineopts_podman="--userns=keep-id"
+DEFAULT_engineopts="$([[ 'docker' == ${REBUILD_ENGINE:-$DEFAULT_engine} ]] && echo $DEFAULT_engineopts_docker || echo $DEFAULT_engineopts_podman)"
+DEFAULT_volumeflags=""
+
+if [ -z "${REBUILD_ENGINE}" ]
```
hboutemy, line 96: there is already a `RB_SHELL` env variable (I can't say well documented...)

can we rename the `REBUILD_*` variables to something like `RB_OCI_*`?

bin/includes/rebuildToolGradle.sh
```diff
@@ -26,27 +26,29 @@ rebuildToolGradle() {
 #      jdkImage="openjdk:17-slim"
 #      ;;
 #  esac
-  jdkImage="gradle:8-jdk${jdk}"
+  jdkImage="docker.io/library/gradle:8-jdk${jdk}"
 
-  info "Rebuilding using Docker image ${jdkImage}"
+  info "Rebuilding using image ${jdkImage}"
```
hboutemy, line 31: `Rebuilding with ${REBUILD_ENGINE}`

jskov:
wilco

jskov:
Created new PR, as I rebased the other branch before realizing it would break the push/PR

PR #161, opened 2024-04-22
Add mada-style gradle plugin

jskov:
These are rebuild instructions for https://github.com/jskov/mada-style-gradle


hboutemy:
reworked the buildspec and rebuilt locally in [6fa3472](https://github.com/jvm-repo-rebuild/reproducible-central/commit/6fa347250f24d4e8af7c443b42678ca8067fbc9b)

no need to produce buildinfo file during build: the rebuild script does the job

PR #163, opened 2024-04-22
Add buildinfo gradle plugin

jskov:
These are rebuild instructions for https://github.com/jskov/buildinfo-gradle/


hboutemy:
as seen in #161 , this buildinfo plugin is not necessary (and despite the description in GitHub, it does not help make a build reproducible: it's just taking fingerprints to easily show if the output is same or different)

jskov:
Huh!

I totally missed that.

Not because the docs were not clear (about what needed to be done for Gradle), mind you.
But because I assumed the build had to produce the buildinfo-file - as other plugins did.

Anyway, I have trimmed the plugin down to what is necessary.
I will submit a new PR for it when ready.

Cheers!

PR #165, opened 2024-05-01
Add Vault-Maven-Plugin

HomeOfTheWizard:
This is the BuildSpec and comparison result for the [vault-maven-plugin](https://github.com/HomeOfTheWizard/vault-maven-plugin)


hboutemy:
notice that buildspec should use the `deploy` profile to rebuild source jar
and in that case, you'd discover that the source jar is not reproducible: it uses an old plugin, you should upgrade it

HomeOfTheWizard:
Well seen! and thanks for pointing out.
I will create another PR to fix that.

PR #166, opened 2024-05-03
Update buildspec for v1.1.6, add new reproducible buildspec for v1.1.7

HomeOfTheWizard:
This is an update for the build info of Vault-Maven-Plugin.
I have updated the buildspec of 1.1.6 and provided the actual result (non reproducible).
Added a new release that is really reproducible (1.1.7), and its buildspec.


hboutemy:
great, thank you

PR #168, opened 2024-05-09
Rebuild information for buildinfo-gradle

jskov:


jskov:
Ah, forgot to purge the obsoleted task.

PR #169, opened 2024-05-18
Add buildinfo-gradle

jskov:


hboutemy:
I read the code of this plugin, and it should definitively not be named "buildinfo" as it does nothing about buildinfo :)

it should be renamed to really match what it does = do some Gradle Reproducible Builds configuration

but this does not prevent it from being reproducible itself :)

jskov:
Yeah, I was just lazy, I guess. You have a good point.
Thanks for merging!

PR #173, opened 2024-05-27
add hibernate-search-parent-7.2.0.Alpha1

marko-bekhta:
Hello 😃 !

This PR is about adding [Hibernate Search](https://github.com/hibernate/hibernate-search) to the list.

With the most recent 7.2.0.Alpha1, the builds should be reproducible. I'm not 100% sure whether I've specified/named everything in the `buildspec` file as you'd expect, so if you have any suggestions or questions, please let me know.

Hibernate Search is a multi-module Maven project publishing multiple artifacts. It does not publish the parent pom starting with 7.0 series, where we started to generate flattened poms for the published jars. Because of that, I'm hoping that using the `hibernate-search-bom` as an artifact id in the spec makes sense while the `buildinfo` is still collected from the root project, specified explicitly as `buildinfo=target/hibernate-search-parent-${version}.buildinfo`

Thanks!


hboutemy:
looks nice, merging, thank you

PR #175, opened 2024-06-03
add json-rpc-0.7.0

sebastian-toepfer:


hboutemy:
FYI, I fixed the buildspec: your build is now proven reproducible :)
and published newer 0.8.0 release

sebastian-toepfer:
@hboutemy thx, yes my idea with the benchmark is not as good as it first looked :(.

PR #178, opened 2024-06-10
add hibernate-search-parent-7.2.0.Alpha2

marko-bekhta:
Hello @hboutemy 😃 

We've released the next alpha of Hibernate Search, and I wasn't sure whether the new versions should be added manually or if that happens automatically ... so I'm opening this PR just in case. (and the build for alpha2 is still reproducible! 😃) Thanks!


hboutemy:
thanks @marko-bekhta 
yes, there is automation to detect new releases: and yes, the new release is still reproducible :) 
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/hibernate/search/hibernate-search-bom/README.md

hboutemy:
notice: I just tried to rebuild with `mvn verify` instead of `mvn package` to check `-sources.jar` and I found sources are not reproducible: 
```
├── org/hibernate/search/backend/lucene/reporting/impl/LuceneSearchHints_$bundle.java
│ @@ -5,15 +5,15 @@
│  import javax.annotation.processing.Generated;
│  import org.hibernate.search.engine.backend.reporting.spi.BackendSearchHints;
│  import java.lang.String;
│  
│  /**
│   * Warning this class consists of generated code.
│   */
│ -@Generated(value = "org.jboss.logging.processor.generator.model.MessageBundleImplementor", date = "2024-06-10T15:35:17+0000")
│ +@Generated(value = "org.jboss.logging.processor.generator.model.MessageBundleImplementor", date = "2024-06-11T06:20:14+0000")
│  public class LuceneSearchHints_$bundle implements LuceneSearchHints, BackendSearchHints, Serializable {
│
```

marko-bekhta:
Hey 😃 

> yes, there is automation to detect new releases: and yes, the new release is still reproducible :)

thanks!

> notice: I just tried to rebuild with mvn verify instead of mvn package to check -sources.jar and I found sources are not reproducible:

hmm 😕, I'll take a look at that. we'll probably have to make a patch for the logging lib to use the `project.build.outputTimestamp`  timestamp... I'll post an update here once we'll figure it out 😃

PR #179, opened 2024-06-23
Gradle plugin buildinfo-gradle renamed to reproducible-gradle

jskov:
Finally got the time to rename my plugin to better reflect what it does :)


hboutemy:
great job, thank you

PR #181, opened 2024-06-28
add scram 3.1 and stringprep 2.2

### LINKED ISSUE #180: Add com.ongres:scram 3.1 and com.ongres:stringprep 2.2
com.ongres:scram 3.1 and com.ongres:stringprep 2.2 versions are reproducible.

jorsol:
Closes #180


hboutemy:
when adding stringprep 2.2, removing 2.0 was not really necessary: having past releases non-fully reproducible is just an info
but let's go :)

Issue #185, opened 2024-07-06
Not able to reproduce rebuild issues in Apache Shiro 2.0.1


lprimak:
Shiro 2.0.1 is indicating a reproducible build issue. However, when I run reproducible build locally, I cannot "reproduce" this issue, and there are no issue reported in the local RB build.

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  07:09 min
[INFO] Finished at: 2024-07-07T00:12:55Z
[INFO] ------------------------------------------------------------------------
dos2unix: converting file target/shiro-root-2.0.1.buildinfo to Unix format...
dos2unix: converting file target/shiro-root-2.0.1.buildcompare to Unix format...

[INFO] rebuild from content/org/apache/shiro/shiro-2.0.1.buildspec
[INFO]   results in content/org/apache/shiro/shiro-root-2.0.1.buildinfo
[INFO] compared to Central Repository content/org/apache/shiro/shiro-root-2.0.1.buildcompare:
    ok=106
[...]

PR #170, opened 2024-05-19
Adds support for using Podman instead of Docker

jskov:
(had to create a new branch, as I rebased my branch by mistake)

These changes allow me to run a (randomly chosen) maven build:

```console
$ ./rebuild.sh content/commons-io/commons-io-2.16.1.buildspec
...
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  24.705 s
[INFO] Finished at: 2024-05-19T18:55:25Z
[INFO] ------------------------------------------------------------------------
dos2unix: converting file target/commons-io-2.16.1.buildinfo to Unix format...
dos2unix: converting file target/commons-io-2.16.1.buildcompare to Unix format...

[INFO] rebuild from content/commons-io/commons-io-2.16.1.buildspec
[INFO]   results in content/commons-io/commons-io-2.16.1.buildinfo
[INFO] compared to Central Repository content/commons-io/commons-io-2.16.1.buildcompare:
    ok=7
    okFiles="commons-io-2.16.1.pom commons-io-2.16.1.jar commons-io-2.16.1-tests.jar commons-io-2.16.1-sources.jar commons-io-2.16.1-test-sources.jar commons-io-2.16.1-cyclonedx.xml commons-io-2.16.1-cyclonedx.json"
    ignored=1
    ignoredFiles=commons-io-2.16.1.spdx.json
```

Let me know if it breaks anything for you/docker.

PR #187, opened 2024-07-27
add JDK 22 support in contatiner mode #185

hboutemy:
Maven DockerHub image name for JDK 22 is different from previous JDK versions => need to add a new rule

while at it, fixed some little pre-existing issues:
- using user name in `docker run -u user:group` instead of user id fails sometimes on Linux
- podman support added in #170 was done only for toolchains case, not base maven image

all that done to rebuild Shiro 2.0.1 in container mode, because surprisingly, aspectj output in container rebuild is ok but not in interactive mode: see #185


gemmellr:
This change seems to break the ability to run the rebuild using podman for me.

I previously ran a rebuild some weeks back and all was well. Today I updated and could not get it to work when doing similar things. Having a look at recent changes and then reverting only this change https://github.com/jvm-repo-rebuild/reproducible-central/commit/9fb4589f6cbab6a40910c980dd552e5338e877d6, it then worked again.

With this in place, I get:
```
INFO] Adding layer to rebuilder container image for local user
/path/to/reproducible-central/bin/includes/mvnBuildDocker.sh: line 167: filename: command not found
[RUN ] podman build --format docker -t -myusername -f /path/to/reproducible-central/bin/docker/tmp/Dockerfile--myusername /path/to/reproducible-central/bin 

Error: tag -myusername: invalid reference format

[FAIL] /========================================================================
[FAIL] |  Unable to build -userid using Dockerfile--myusername 
[FAIL] \========================================================================
[FAIL]  ---------->>> PROCESS WAS ABORTED <<<---------- 
```

hboutemy:
replaced `filename` with `basename`, that is more portable

gemmellr:
Great, thanks!

PR #188, opened 2024-07-30
Add ActiveMQ Artemis 2.36.0

gemmellr:
Add ActiveMQ Artemis 2.36.0

Note: I used "_clean install"_ rather than "_clean package_" as the 2.34.0 config did. I did this because using "package" whilst verifying against the staging repo before release would otherwise incorrectly say a few modules sources jars created by the shade plugin are not reproducable when they actually should be. This happens because the shade plugin needs the sources jars from other modules in the build and Maven doesnt have access to the staging repo for its regular repo usage, only the comparisons, so when checking agaisnt the staging repo it doesnt find the sources if they arent installed and then just skips their inclusion and generates the incomplete shaded sources jar. It would find them with "package" once released to Maven Central.

That is also what I expect happened with the Artemis 2.34.0 addition made previously since its commit https://github.com/jvm-repo-rebuild/reproducible-central/commit/e92ca27d3e0e09e0aed23ba1db6191e324597db3 is dated the day before 2.34.0 was actually released, presumably meaning it was run against the staging repo. Running the same 2.34.0 config now that 2.34.0 was released to Maven Central would give a different result for the 3 sources jar variations since Maven would now be able to find the other modules sources jars to construct the shaded one properly.

(Its also possible to resolve the single [dependency-reduced-]pom variation for 2.34.0 by changing the profiles used and maven version; I found there are at least 3 repeatable outputs depending on the combinations used. I have updated things to avoid that issue for 2.36.0 via https://github.com/apache/activemq-artemis/commit/6fe1997aec7fc05ccc6af77c6ceec877a1eeba13 so that 2.36.0 isnt affected by that potential variability so 2.36.0 doesnt need any such consideration, and fixed a final issue with the release profiles 'website' module via https://github.com/apache/activemq-artemis/commit/ae8fd6dbdbae2653194512b97484a27440a38f6d)


hboutemy:
`install` is definitively a no go: comparison done by `artifact:compare` happens against what is installed, then does not download the reference

> incorrectly say a few modules sources jars created by the shade plugin are not reproducable when they actually should be
if the result is unexpectedly different, and we can measure the difference with diffoscope, that's a fact. Using `install` is lying :)

if you want to use `verify` instead of `package`, you can

if you think the build definitively requires `install`, we should review which plugin requires install because it should not

we'll need to work together to see what changes in the buildspec are ok and improve the result (because yes, the buildspec may be improved)

gemmellr:
Ah, interesting on the install vs compare.

I'm not sure that `verify` would make a difference (but havent tried) based on what appears to be happening. Agaisnt the staging repo, the shade plugin indicated it couldnt get the sources for an earlier module in the build (artemis-commons, which is itself also shading) whilst it is creating a sources jar for a later module that shades the earlier module. This happens in the 3 modules (artemis-core-client-all, artemis-jms-client-all, artemis-jakarta-client-all, each doing much the same thing and creating an ugly uber jar) that incorrectly get marked non-reproducible since the shade plugin just logs a warning and continues, so those 3 'generated' sources jars are then end up incomplete as it didnt include some stuff it should which it does during the release as that installs.

What might make a difference rather than using "install" is giving Maven access to the reference repo in general (rather than just for artifact:compare) in addition to Central.  As I say, if you run that same 2.34.0 check again now youll find it gives a difference result, the 3 jar differences will go away as same thing with "package" that failed agaisnt the staging repo will work fine when running against Central now that the release has happened, because it is then able to get the earlier module's sources jar (from central this time) much as it would during the artemis release process (but locally then, because they are installed).

gemmellr:
Ah, so trying things again with "package" I see the output actually changed in a way I didnt expect (less artifacts, specifically sources jars.....) and so it seems likely "verify" could indeed behave differently against the staging repo (harder to check as the staging repo is gone now) because the build oddly has maven-source-plugin config tied to the _verify_ phase for some reason. I may look into why that is the case and perhaps changing that in future, but for now I'll give things a run with verify.

gemmellr:
I updated the buildspec to use "verify" instead of install, then deleted the buildcompare and buildinfo, cleaned my git repo, deleted the buildcache, and deleted the containers. Running everything afresh, I then got the same buildcompare and buildinfo output as before.

I could see the shade bit working / not warning, and the compare step downloading things from Central. So seems like "verify" should work against the staging repo too, and the ultimate issue is the weird configuration tying the source plugin to the verify phase. I'll look into why that was done and perhaps changing it to something more usual in future, thanks for the pointer!

hboutemy:
ok, I tried to rebuild your artemis 2.36.0 buildpsec and got the same success as you: great, I'll merge the PR

on issues from previous releases:
- shaded sources and requirement to install (faked by artifact downloaded from central): it seems upgrade of maven-shade-plugin from 3.5.2 to 3.6.0 solved the issue, as the dependency should be used from reactor, not from local repository nor downloaded
- I don't have any idea for now why the dependency reduced pom is also more stable

let's celebrate the good news: 2.36.0 does not have any problems any more, neither about reproducibility nor unwanted install prerequisite
on past releases, past is past, this is why plugins upgrades are useful: they benefit from fixes even if we don't precisely know we benefit from it

hboutemy:
notice: I just found https://issues.apache.org/jira/browse/ARTEMIS-4822
linked to it in 2.25.0 release
thanks a lot for your hard work on this topic: now I know success is yours :)

gemmellr:
When I noticed the pom disparity with your 2.34.0 build (and what I got locally) with what was released, I asked the person that released it what maven they used (an earlier 3.8.x, I forget which), and tried it and found it gave a different result. Realising that, I then tried it with a variety of active profiles and different maven version combinations and found multiple seperate 'repeatable' outputs in the combinations I tried. From memory I think the maven version aspect of it changed with Maven 3.9.2, whilst the buildspec used 3.9.3 and the release used a 3.8.x. I reworked the pom so that the profile usage in that pom is only a simple property toggle and the stuff that varied is all outside the profiles. That seems to be happy with any combination of maven and profiles used and always generated the same output regardless what was used.

PR #194, opened 2024-08-19
Add ActiveMQ Artemis 2.37.0

gemmellr:
This is a draft for now, as Artemis 2.37.0 is still under vote. The rebuild was verified against the staging repo, and I decided just to open the PR to make aware it was coming. I will update things as needed and convert when a vote has passed.

The buildspec differs slightly from the one for 2.36.0 (#188), changing things back to use the typical "package" phase rather than "verify" during the rebuild, following adjustments in [ARTEMIS-4980](https://issues.apache.org/jira/browse/ARTEMIS-4980).


gemmellr:
Artemis 2.37.0 was released, so just rebased the previous runs output (verified against staging repo) against current master.

gemmellr:
> great, thanks FYI, I generate new release buildspec easily with `bin/add-new-release.sh content/org/apache/activemq/artemis/artemis-2.36.0.buildspec 2.37.0`:

I realised; that is actually exactly how I did it too originally. Which failed as I was running against staging and forgot to add the "staging" arg, which I then did and reran it :)

I wanted to change it back to the usual "package" you had been using, now that the artemis build actually facilitates doing so, so I then updated the buildspec and re-ran again.

> I did it, it rebuilt and found 156 artifact ok, vs yours with 152 I suppose this is due to my rebuild using phase `verify` while yours is stopping a little bit earlier at `package` phase
> 

Thats correct, I noticed that the same. The 4 added checksums are generated by an execution like the one added by the apache parent pom, which only happens at "verify" stage. I thought it was nicer using "package" without the checksums, since checksums are neither created or validated for any of the other modules, or most other apache projects that you usually seem to rebuild with "package".

> consequence is that I'll commit my rebuild instead of this PR, sorry
> 

No worries, I dont personally care which approach it uses. Verify is closeer to what the release does. I actually just thought you would prefer to have the buildspec back at "package" without the checksums, to be consistent with the rebuilds of many other projects youve use "package" for (e.g ActiveMQ 'classic').

> notice that for testing during VOTE from staging area, `bin/add-new-release.sh content/org/apache/activemq/artemis/artemis-2.36.0.buildspec 2.37.0 staging` works also

Yep, that is actually what I did, then modified the buildspec to "package" and re-ran, to remove the seeming outlying prior use of "verify".

hboutemy:
'4 added checksums"? checksums? uh?
looking at it, I found `apache-artemis-2.37.0-source-release.zip.sha512 apache-artemis-2.37.0-source-release.tar.gz.sha512 apache-artemis-2.37.0-bin.zip.sha512`

ok, now I see https://repo.maven.apache.org/maven2/org/apache/activemq/apache-artemis/2.37.0/

and I see the root cause https://github.com/apache/maven-apache-parent/blob/master/pom.xml#L486

I did not expect that edge case....

thanks for pointing me to that, yet another small build unusual detail that Reproducible Builds checks permits to point to

Issue #190, opened 2024-08-02
The component README.md can have broken links to modules


gemmellr:
The README.md for each component being rebuilt lists a count of how many modules it defines, and has an expandable set of links to the modules details on _central.sonatype.com_. These links are all use the latest rebuilt version in their URL GAV details, but seem to be for the aggregated set of modules from _all_ previously rebuilt releases. This means the links 404 if those modules no longer exist in the current release, e.g due to something being renamed or removed previously.

For example, see the [artemis README.md](https://github.com/jvm-repo-rebuild/reproducible-central/blob/d75f36910acdb08e5bb6a712a71ece4e47d48d12/content/org/apache/activemq/artemis/README.md) following addition of 2.36.0. It indicates "This project defines 218 modules:" and expands to links for them, all at version 2.36.0. The first 150+ links all 404 because they are for examples that dont exist in recent versions of the build, having previously moved out to their own repository/build (and no longer being deployed). There are multiple other later links that similarly 404 for modules that were either renamed or simply deleted in the past.

Feels like the module links should only exist for the latest version modules (or else use the latest version they actually existed for, so they are functional links)?

hboutemy:
good catch, good idea
code is https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/bin/update-reproducibility-summary.sh#L125

gemmellr:
I have raised a draft PR #195 with a potential update to the script, and the effects of running it.

gemmellr:
I updated and removed draft status on the PR last week. Thought I had mentioned it here at the time, but seems I only did it on the PR itself.

hboutemy:
see result https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/README.md

PR #195, opened 2024-08-19
make README links use modules from newest release with .buildinfo

gemmellr:
Updates the `bin/update-reproducibility-summary.sh` script so that the README file overview links only include modules determined from the newest _released_ version .buildinfo file.

I have made this PR a draft for now, as it currently incorporates my changes from #194, which is similarly a draft. I did that because I wanted to verify that it still selected 2.36.0 modules for the README links for artemis, since 2.37.0 is still not released. I'll update the PR to address that later, either removing those changes, or once it is released and added.

I also wasnt sure whether this PR should include _just_ the script changes, or _all_ of the updates it causes. I thought it worth raising the draft PR with them all included for now so that the overall effect can be observed for review, even if the non-script updates should then be removed and left for the CI refresh to apply after merge.

### DIFF
--- a/bin/update-reproducibility-summary.sh
+++ b/bin/update-reproducibility-summary.sh
@@ -86,6 +86,17 @@ do
       break
     fi
   done
+  # detect newest version that has a buildinfo file
+  newestBuildinfoVersion=
+  for version in $($tac "${metadata}" | grep 'version>' | cut -d '>' -f 2 | cut -d '<' -f 1)
+  do
+    if [ -n "$(ls $dir | grep "\-${version}\.buildinfo")" ]
+    then
+      newestBuildinfoVersion="$version"
+      break
+    fi
+  done
+
   latestVersion="$(cat "${metadata}" | grep 'latest>' | cut -d '>' -f 2 | cut -d '<' -f 1)"
   lastUpdated="$(cat "${metadata}" | grep 'lastUpdated>' | cut -d '>' -f 2 | cut -d '<' -f 1)"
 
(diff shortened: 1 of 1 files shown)


gemmellr:
I did a sampling of a a variety of the 'removed links' across a spread of components, all the removed links I tried 404'd  originally.

gemmellr:
I stripped out the previous runs readme updates (they can still be seen in the old commit https://github.com/jvm-repo-rebuild/reproducible-central/commit/66dfa6a48fd6a88f15e3005b60171ab4a7d7090f for review), leaving just the `bin/update-reproducibility-summary.sh` script changes themselves, and rebased against current master. If you would prefer the PR make all the readme updates too let me know.

bin/update-reproducibility-summary.sh
```diff
@@ -132,15 +143,15 @@ do
...
         if [ $projectGa -gt 1 ]
         then
           echo "<details><summary>This project defines $projectGa modules:</summary>" >> ${projectReadme}
           echo >> ${projectReadme}
-          for ga in $(cat $dir/*.buildinfo | grep coordinates | cut -d = -f 2 | sort -u)
+          for ga in $(cat $dir/*-${newestBuildinfoVersion}\.buildinfo | grep coordinates | cut -d = -f 2 | sort -u)
           do
             gaDir=$(echo "$ga" | sed -e 's_:_/_')
-            echo "* [$ga](https://central.sonatype.com/artifact/${gaDir}/${version})" >> ${projectReadme}
+            echo "* [$ga](https://central.sonatype.com/artifact/${gaDir}/${newestBuildinfoVersion})" >> ${projectReadme}
```
hboutemy, line 154: IIUC, this will apply the same version for each and every module: the issue is that some modules disappeared, some were added, then there is no single value matching all modules, it would have to be per-module

checking again, as of now, I can just use the `overview` url, without any version: it matches what we need IMHO, it was just not there when the feature was added

I'll switch to the non-versioned url

gemmellr:
> IIUC, this will apply the same version for each and every module: the issue is that some modules disappeared, some were added, then there is no single value matching all modules, it would have to be per-module
> 
> checking again, as of now, I can just use the `overview` url, without any version: it matches what we need IMHO, it was just not there when the feature was added
> 
> I'll switch to the non-versioned url

Yep, the issue I noticed and raised #190 for was the readme linking with newest-version to historic modules from old versions which no longer exist for whatever reason, and so 404'ing with the newest-version-specific links being created.

The PR idea was to use the known set of modules defined by the most recent actual rebuild, which are guaranteed to exist in that version, and to me seems the best actual fit for a "This project defines <count> modules:" overview. Linking to a variety of versions, or the no-version overview, where you need to look at every single link to identify whether it exists anymore, didnt seem as nice a fit to me.

It feels more like it would be better described different, e.g "This project has, over time, defined <count> modules:". Though it could itself still be somewhat innacurate, since it will only know about modules that have been rebuilt at some point, and few have had every version rebuilt, so its in some cases likely still not the actual list of modules the project has historically defined. So maybe "The project has, over time, had <count> modules rebuilt:" would better describe what is being shown.

gemmellr:
In any case, this PR is not being used, so closing out.

PR #200, opened 2024-09-14
Jdk22 tool chains

nielsbasjes:
Open problems:

An image with this
```
tool=mvn-3.9.9
jdk=21
toolchains="8|17|21|22"
```
should have jdk 22 in the toolchains and jdk 21 as the default.
Both are not the case yet.

```
# update-java-alternatives -l 
java-1.11.0-openjdk-amd64      1111       /usr/lib/jvm/java-1.11.0-openjdk-amd64
java-1.17.0-openjdk-amd64      1711       /usr/lib/jvm/java-1.17.0-openjdk-amd64
java-1.21.0-openjdk-amd64      2111       /usr/lib/jvm/java-1.21.0-openjdk-amd64
java-1.8.0-openjdk-amd64       1081       /usr/lib/jvm/java-1.8.0-openjdk-amd64
```

```
# update-java-alternatives --set "$(update-java-alternatives -l | fgrep "21" | sed 's@ \+@ @g' | cut -d' ' -f1)" 
update-alternatives: error: no alternatives for jaotc
```

### DIFF
--- a/bin/docker/Dockerfile.environment.template
+++ b/bin/docker/Dockerfile.environment.template
@@ -29,20 +29,20 @@ RUN chmod 755 /scripts/*.sh
 ###
 ADD docker/mvn /usr/local/bin
 RUN chmod 755 /usr/local/bin/mvn
-ENV PATH /usr/local/bin/:${PATH}
+ENV PATH="/usr/local/bin/:${PATH}"
 
 # Avoid out of memory errors in builds
-ENV MAVEN_OPTS -Xms256m -Xmx512m @@MAVEN_OPTS@@
+ENV MAVEN_OPTS="-Xms256m -Xmx512m @@MAVEN_OPTS@@"
 
 ###
 # Better reproducibility: Make the desired locale explicit
 # Set the locale ( see https://stackoverflow.com/a/28406007/114196 )
 ###
-ENV BUILD_LOCALE @@BUILD_LOCALE@@
+ENV BUILD_LOCALE=@@BUILD_LOCALE@@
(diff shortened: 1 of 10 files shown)


nielsbasjes:
Trying to fix the update-alternatives ...
```
cd /usr/lib/jvm/ && ln -s /opt/java/openjdk/ java-22-openjdk-amd64
cd /usr/lib/jvm/ && ln -s java-22-openjdk-amd64 java-1.22.0-openjdk-amd64 
cd /usr/lib/jvm/ && sed 's@21@22@g' .java-1.21.0-openjdk-amd64.jinfo > .java-1.22.0-openjdk-amd64.jinfo 
```

All the things must be installed at the `update-alternatives` end too.

Seems like the java 11 compiler (javac) is not installed ?!?! Only the JRE?

nielsbasjes:
@hboutemy This is my approach to having toolchains available in a situation where a non-LTS version of Java is used. Because the non-LTS is installed "manually" it took some effort to get it work together nicely if a toolchain is needed (essentially making update-java-alternatives work on that JDK also).

The current solution requires jdk 21 installed if you need toolchains for jdk 22 because this solution direction copies some files and uses them to setup 22 correctly.

I have also added the new config for the next release of yauaa (should be out soon) which needs all of this.

Let me know what you think.

nielsbasjes:
Yauaa 7.28.1 has been released

nielsbasjes:
I'm rewriting the jinfo file part. I think I can do better than depend on java 21.

PR #318, opened 2024-10-02
add glassfish-parent 7.0.18

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/main/README.md


hboutemy:
244 issues instead of 11 in previous release: need to investigate the diffoscope output to see if these are more tied to a buildspec weakness of if there is really a reproducibility regression in the project release

PR #359, opened 2024-10-08
fix: improve crlf support for gradle

iProdigy:
Gradle produces pom files with line endings congruent to the system line separator property. See https://github.com/gradle/gradle/issues/19866 and https://discuss.gradle.org/t/poms-are-written-with-the-default-line-ending-of-the-platform/1667 for example

If a project was initially built on windows, when we build with linux, none of the pom files are reproduced due to clashing line endings

This PR converts LF to CRLF for the generated pom files when running gradle in crlf* mode


hboutemy:
great addition: thank you

PR #416, opened 2024-10-16
add cyclonedx-core-java 9.1.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/cyclonedx/cyclonedx-core-java/README.md


hboutemy:
require JDK 8 Azul distribution instead of OpenJDK, because Azul backported additional hash algorithms

Issue #483, opened 2024-10-23
add buildspec workdir configuration


hboutemy:
see https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/opentelemetry/instrumentation/opentelemetry-instrumentation-api-2.0.0.diffoscope
the only issue on 1000+ artifacts is just the full path to source code

if we could configure the path in the container to configure the same as the original builder, we could solve some issues that are not really "not feasible": workdir is just an environment requirement that require some tricks to be easy to configure

hboutemy:
it works!

PR #485, opened 2024-10-23
add opentelemetry-instrumentation-api 1.30.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/opentelemetry/instrumentation/README.md


hboutemy:
contains Docker build scripts updates for #483

PR #626, opened 2024-11-08
Only run scheduled workflows on jvm-repo-rebuild/reproducible-central

jskov:
I have tested that it successfully disables workflow runs in my clone.

But you should double check that it still works on your repo :)


hboutemy:
good idea, thank you

PR #657, opened 2024-11-14
add orc 2.0.3

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/orc/README.md


hboutemy:
GHA rebuild done on `native-linux-x86_64`
reference build done on `native-osx-aarch64`

ideally, I should improve the GHA setup to use an OSX CI node => I'll open an issue to track this dream

I'll probably build on my own machine and override the result provided by GHA

PR #767, opened 2024-11-25
add nifi 2.0.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/nifi/nifi/README.md


content/org/apache/nifi/nifi/nifi-2.0.0.buildspec
```diff
@@ -0,0 +1,19 @@
...
+gitRepo=https://github.com/apache/${artifactId}.git
+gitTag=rel/${artifactId}-${version}
+
+tool=mvn-3.9.8
+jdk=21
+newline=lf
+locale=en_US
+umask=002
+
+command="mvn -Papache-release,include-grpc,include-ranger clean package -DskipTests -Dmaven.javadoc.skip -Dgpg.skip"
```
exceptionfactory, line 15: @hboutemy For reference in the next release version, NiFi now uses the Maven Wrapper command `mvnw`, which should improve reproducibility by ensuring the same version of Maven. In addition, the `include-grpc` and `include-ranger` profiles are no longer part of the NiFi 2.0 build process, so they can be removed. Thanks for continuing to track reproducibility for NiFi.

hboutemy: thanks for the hint: testing in #836
adding new releases is easy by copying previous recipe, getting your help on adapting is very useful
we'll now need to find each small improvement we can do on the remaining RB issues: for example, I now remember I use your project as an example https://github.com/apache/nifi/pull/6500/files#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8 but the result is incomplete...

one nice thing to do at project level is to share rebuild result during vote process: don't fail the vote on remaining issues, but it's just an opportunity to share and celebrate improvements

PR #768, opened 2024-11-25
fix: use https to clone `atom-jaxb`

algomaster99:
I was running `./rebuild.sh` on a server and I don't have my ssh credentials configured so the clone failed. Proposing this to ensure no breakage.


algomaster99:
The CI seem to fail because the build is not reproducible.

algomaster99:
> your change on the buildspec did not change anything in the output = something unexpected

In theory, that is true. But here the URL to source repo is the same. I only changed the protocol. I was not expecting anything to change as well. :sweat_smile:

PR #836, opened 2024-11-30
update NiFi 2.0.0 buildspec as suggested by project

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/nifi/nifi/README.md


hboutemy:
uh, it's bringing more issues than the original buildspec...
let's see for the next release: I'll keep this PR open to not forget about it

PR #875, opened 2024-12-03
add avaje-jex-parent 3.0-RC8

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/avaje/jex/README.md


hboutemy:
based on previous experience of other RCs, there are 2 commitss: for RC8, only the second one has been committed
then no surprise RC8 builds output that says RC7, this is the missing commit

PR #914, opened 2024-12-07
add yupiik-tools-maven-plugin-parent 1.2.5

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/yupiik/maven/plugin/README.md


hboutemy:
same issues as previous release

PR #1038, opened 2024-12-18
separate ratis and ratis-thirdparty

adoroszlai:
Apache Ratis releases both Ratis and Ratis Third-party with `org.apache.ratis` group.  Splitting to subdirs to get separate README, badge, etc.


hboutemy:
I try to keep 1 buildspec per PR, because each time there is a build from CI behind it
in general, I start with 1 release for each project, then once it it stabilized, I see how to cover releases history from the stabilized project

yes, clearly, these are 2 projects: each requires a directory in Reproducible Central

please do 1 PR for moving ratis third party, and one PR for one release for ratis (in general the latest, which is hte most reproducible)

you've been hit by one of the key organizational questions = what is the "right" directory for a project. Clearly it is based on groupId, sometimes it is also based on a choice of one artifactId or any other choice that is more human friendly...

hboutemy:
thank you for helping: it's good to see someone trying as seriously, then discovering issues I discovered before :)

adoroszlai:
I guess `ratis` can stay where it is now.  `logback` has similar layout.

PR #1256, opened 2024-12-29
add maven-lockfile 5.3.5

LogFlames:
https://github.com/chains-project/maven-lockfile/releases/tag/v5.3.5

I had some trouble with our release CI, accidentally causing a bunch of nearly-empty releases but are now up and running again.

How is the best practice to add a buildspec for a new version? Is the preferred way to create a PR or mail the mailing list?Or did it stop auto-detecting new releases for some other reason?

I also want to apologize for any manual labor the empty-releases potentially caused!


hboutemy:
thanks for this PR

when latest release is 100% reproducible, I systematically detect and rebuild new releases; When it's near, I do it eventually, but when I can't build, it goes down in my priority list...

This PR is the right way to give me a good reason to work again on new releases: ideally, diving into the last non-reproducible bits would be great.

I'll also do the job for previous releases, to see what's the best rebuild we can do

PR #1363, opened 2025-01-08
fix cache-api 0.7.0

iProdigy:
Follow up to #1320

Resolves reproducibility issue for `cache-kotlin-0.7.0-sources.jar` caused by https://youtrack.jetbrains.com/issue/KT-54207/Kotlin-has-two-sources-tasks-kotlinSourcesJar-and-sourcesJar-that-archives-sources-to-the-same-artifact

I'll edit the relevant build.gradle file for the next version to supplant this buildspec hack à la https://github.com/micrometer-metrics/micrometer/commit/d9d6599ba742df8e4864731f6c187f1bb792153c


iProdigy:
aside: CI failed to add you as a reviewer

```
error fetching organization teams: GraphQL: Resource not accessible by integration (organization.teams)
```

which is caused by https://github.com/cli/cli/issues/4844

hboutemy:
thanks for digging and fixing: i don't precisely get why you had to exclude a task, but I suppose you'll improve you build for future releases to not require this, isn't it?

for my script failing to add reviewer, thanks for the feedback: I could not easily test, I'll have to just let the issue open without any sign...

iProdigy:
Yeah the build for the next release has been fixed, so this manual task exclusion will be unnecessary

it's also a perplexing bug to me: normal jar assembly has the tasks in the right order but publishing to maven local somehow caused the kotlinSourcesJar task to run after the normal sourcesJar task, overwriting the "good" jar with one that included a `main` directory - so I've disabled this unwanted task from my build going forward

PR #1546, opened 2025-02-06
add migratedb 1.4.1

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/de/unentscheidbar/migratedb/README.md


hboutemy:
JDK 17.0.14 creates different bytecode for 1 class: using 17.0.13 is ok

PR #1547, opened 2025-02-06
add langchain4j 1.0.0-beta1

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/dev/langchain4j/README.md


hboutemy:
https://github.com/langchain4j/langchain4j/issues/2514

hboutemy:
it's better, but still not building https://github.com/jvm-repo-rebuild/reproducible-central/actions/runs/13349163865/job/37283398097?pr=1547

hboutemy:
beta2 is ok, let's just say that beta1 cannot be built, past is past...

PR #1562, opened 2025-02-08
add gemLibPki 3.1.1

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/de/gematik/pki/gemLibPki/README.md


hboutemy:
when built with JDK 17.0.14 (as the GHA rebuilder uses), jar is different, with different .class

I had to do a local build with JDK 17.0.x with x < 14: tested with 13 and others, it's ok

:( we're back to something I saw in the past with some JDK 8, but not again until now

PR #1675, opened 2025-02-16
add convex 0.8.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/world/convex/README.md


hboutemy:
missing Git tag https://github.com/Convex-Dev/convex/issues/522

PR #1676, opened 2025-02-17
Don't remove `build` directories in the Gradle rebuild

marko-bekhta:
Hey 👋🏻 😃,

I was looking into adding one more Hibernate project to the list, but encountered a problem with this Gradle rebuild script (among some other things). 
We have some packages called `build` (e.g. https://github.com/hibernate/hibernate-orm/tree/main/local-build-plugins/src/main/java/org/hibernate/build) and running that `rm` leads to build failures as parts of the code are missing. 

Or was there some specific reason for this `rm` that I'm missing?

### DIFF
--- a/bin/includes/rebuildToolGradle.sh
+++ b/bin/includes/rebuildToolGradle.sh
@@ -35,7 +35,6 @@ rebuildToolGradle() {
   [ -d userhome/.m2 ] || mkdir -p userhome/.m2
   [ -d userhome/.gradle ] || mkdir -p userhome/.gradle
   [ -d $PWD/.bnd ] || mkdir -p $PWD/.bnd
-  find . -name build -exec \rm -rf {} \;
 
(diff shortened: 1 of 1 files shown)


hboutemy:
in fact, I'm not a Gradle expert: I just did cleanup of this `build` directory that was a pure output in every other project I rebuilt until now

if you're confident that cleaning up is not necessary, we can try not cleaning up: I just need someone who understand Gradle sufficiently to do updates that respect every other project that will be built with same scripts

marko-bekhta:
hmmm, yeah, I am not a Greadle expert either 😃. I'd expect that having `./gradlew clean` to do the trick (https://docs.gradle.org/current/userguide/java_plugin.html#sec:java_tasks) + adding the `--no-build-cache` to have a complete build from scratch...

hboutemy:
ok, let's consider we'll add clean to buildspec if necessary: ok, I'll merge this PR :)

PR #1817, opened 2025-03-03
fix: log correct path for reference and rebuild artifacts

### LINKED ISSUE #1779: Incorrect paths for diffoscope comparison when Maven build not from root directory
When a rebuild is executed and there are bit-by-bit differences, the script outputs paths for manually execution diffoscope command.

https://github.com/jvm-repo-rebuild/reproducible-central/blob/0e1666340ff6de8d7647f51d45b308eb8bafa829/content/org/apache/orc/orc-1.8.3.buildcompare#L10-L16

However, if you `cd content/org/apache/orc/buildcache/orc` and then run these `diffoscope` commands, you get file does not exist error.
This happens because the root of this project (or where pom.xml) is at [`orc/java`](https://github.com/apache/orc/tree/v1.8.3/java).
The solution to correction of these comments would be if all of these commands also respect `-f java/pom.xml`.

algomaster99:
Fixes #1779 

I chose to go with fixing at `./rebuild.sh` level rather than `maven-artifact-plugin` as I wanted to edit only these two information:
```
[INFO] build available in content/org/apache/orc/buildcache/orc/java, where you can execute diffoscope
[INFO] To see every differences between current rebuild and reference, run:
[INFO]     ./build_diffoscope.sh content/org/apache/orc/orc-1.8.3.buildcompare buildcache/orc/java
```

### DIFF
--- a/bin/includes/displayResult.sh
+++ b/bin/includes/displayResult.sh
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 
 displayResult() {
-  local buildcompare version ok ko okFiles koFiles reference_java_version reference_os_name
+  local pomDirectory="$1"
+  local buildcompare version ok ko okFiles koFiles reference_java_version reference_os_name pomDirectoryInBuildcache
   buildcompare="$(dirname "${buildinfo}")/$(basename ${buildinfo} .buildinfo).buildcompare"
 
   info "rebuild from \033[1m${buildspec}\033[0m"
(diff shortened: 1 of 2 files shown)


hboutemy:
tested on Mac: it does not work :(

```
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

hboutemy:
works now, thanks

PR #1897, opened 2025-03-12
add artemis-pom 2.40.0

clebertsuconic:


hboutemy:
Artemis is already covered at Reproducible Central https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/README.md
and 2.40.0 is not yet available at Maven Central https://central.sonatype.com/artifact/org.apache.activemq/artemis-pom/versions

what are you trying to do?

clebertsuconic:
@hboutemy I am releasing 2.40.0 and I thought the expected step was to send a pull request here when the release is being done.


I will close this now.. and I will talk to my coworker Robbie who did the integration on this.

Sorry for the confusion.

hboutemy:
no problem: all this is quite new
you can test *for yourself* during ASF staging by adding `staging` tothe add and rebuild commands: but it is valid only for yourself, as Git will only store final rebuilds for Maven Central content, not transient content from ASF staging

PR #1966, opened 2025-03-18
add nifi-api 2.0.0

hboutemy:


hboutemy:
was previously part of NiFi

PR #2203, opened 2025-04-10
add hibernate-models 0.9.5

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/hibernate/models/hibernate-models/README.md


hboutemy:
i suppose this is a difference between JDK 17.0.14 and 17.0.13 and less
but I don't see how to check what is actually used by a Gradle build: help welcome

PR #2338, opened 2025-04-21
Remove oras spdx version workarround

jonesbusy:
Version issue was due to https://github.com/spdx/spdx-java-v3jsonld-store/issues/18#event-17323721720
Still remains the SPDX JSON is not reproducible, so ignored for now

Tested with 

```
./rebuild.sh content/land/oras/oras-java-sdk/oras-java-sdk-0.2.6.buildspec
```


jonesbusy:
@hboutemy Hi

Not sure if the correct approach but the build fail

```

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	PRcomment.txt
	time.txt
``` 

Not sure the reason behind this

jonesbusy:
Thanks. No need for older releases due to issue on SPDX plugin

PR #2422, opened 2025-04-28
add avaje-inject-parent 11.5-RC4

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/avaje/inject/README.md


hboutemy:
missing commits for release https://github.com/avaje/avaje-inject/issues/814

PR #2554, opened 2025-05-13
add glassfish-parent 7.0.24

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/main/README.md


hboutemy:
reference was build with JDK 17.0.14, rebuilder using 17.0.15 brings some differences...

PR #2634, opened 2025-05-19
add org.apache.sling.distribution.journal.messages 0.5.12

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/sling/org.apache.sling.distribution.journal.messages/README.md


hboutemy:
JDK 11.0.27 (used by GHA) seems to produce different bytecode from JDK 11.0.25 (used by local rebuilder)
same as #2636

PR #2636, opened 2025-05-19
add slingfeature-maven-plugin 1.9.2

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/sling/slingfeature-maven-plugin/README.md


hboutemy:
JDK 11.0.27 (used by GHA) seems to produce different bytecode from JDK 11.0.25 (used by local rebuilder)

PR #2713, opened 2025-05-24
add klass 0.5.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/cool/klass/README.md


hboutemy:
GHA build done with JDK 17.0.15
local build done with JDK 17.0.13

PR #2917, opened 2025-06-18
add maven-turbo-builder 0.3

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/github/seregamorph/maven-turbo-builder/README.md


hboutemy:
ok, this one is not RB-enabled

PR #2968, opened 2025-06-23
add org.apache.sling.resourceresolver 1.12.10

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/sling/org.apache.sling.resourceresolver/README.md


hboutemy:
filesystem used for the reference build puts executable flag on each and every file, be it source from Git or output of the build
(probably a Windows implementation, that does not know what an executable flag is...)

so in theory reproducible, bu de-facto requires a very specific setup to get these executable files everywhere

hboutemy:
perfect case for testing new stabilize feature #2983

PR #2976, opened 2025-06-25
Update the formulation of the objectives of the project

christian-stj:
Hi! I was planning to cite these goals in my thesis but the grammar was a bit off. Is this an acceptable rewrite? @hboutemy


hboutemy:
it is good, my frenglish, isn't it? ;)

thanks for your help, happy to accept this improvement

PR #2993, opened 2025-06-26
add maven-pmd-plugin 3.27.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/maven/plugins/maven-pmd-plugin/README.md


hboutemy:
TBD: Maven 4.0.0-rc-3 seems to behave differently from previous rc-2 regarding build vs consumer POM

PR #3016, opened 2025-06-30
add opentelemetry-instrumentation-api 2.17.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/opentelemetry/instrumentation/README.md


algomaster99:
@hboutemy here the build fails because `stabilize` cannot work with JSON files hence it exists with 1. Should we also include this case under `stabilize_ko`?

hboutemy:
yes, feature now added, I'll rebase

PR #3027, opened 2025-07-01
add openfasttrace 4.2.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/itsallcode/openfasttrace/README.md


hboutemy:
requires https://github.com/apache/maven-artifact-plugin/issues/173

PR #3076, opened 2025-07-07
add flexmonster 2.9.107

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/webjars/flexmonster/README.md


hboutemy:
no Git tag

PR #3117, opened 2025-07-11
add mybatis-spring-boot 3.0.5

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/mybatis/spring/boot/mybatis-spring-boot/README.md


hboutemy:
japanese directory and file names are causing issues to lf to crlf conversion script
https://github.com/mybatis/spring-boot-starter/tree/master/mybatis-spring-boot-autoconfigure/src/test/java/org/mybatis/spring/boot/autoconfigure

PR #3119, opened 2025-07-11
add log4j 2.25.1

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/logging/log4j/log4j/README.md


hboutemy:
@ppkarwasz any idea what in the buildspec is causing the subtle differences in output (see diffoscope)

ppkarwasz:
Hi @hboutemy,

Can you try adding the `deploy` profile?

```sh
mvn -Pdeploy,release package
```

ppkarwasz:
We've encountered a few reproducibility issues while building Log4j `2.25.1`:

* The aggregate SBOM file (`log4j-bom-2.25.1-cyclonedx.xml`) occasionally reorders the `JSpecify` dependency, shifting it between two positions. I noticed this also occurred in `2.25.0`, as you had already flagged. I'm not yet sure what causes this nondeterminism—I plan to investigate and will report back once I understand the root cause.

* The Gradle Module Metadata (GMM) Maven Plugin includes a `createdBy` key containing the Maven version, which affects reproducibility. Fortunately, this was quickly addressed by the author in gradlex-org/gradle-module-metadata-maven-plugin#43.

* Lastly, while the [`deploy` profile in `logging-parent`](https://github.com/apache/logging-parent/blob/fe3d4172ebe32b7c962909a56c372a475c7de16c/pom.xml#L1309) shouldn't affect packaging, enabling it appears to change the execution order of the GMM and Source plugins. As a result, the comment inserted by the GMM plugin in `pom.xml` may or may not end up in the `-sources.jar` archive.

hboutemy:
yes, I saw the first and the third issue: for the second, selecting expected Maven version is easy, so not a strong problem (it does not "affect reproducibility" but affects ease of reproducing as rebuild environment is just more strict)

on the third, looking at build log
```
[INFO] --- jar:3.4.2:jar (default-jar) @ log4j-api-test ---
[INFO] Building jar: /var/maven/app/log4j-api-test/target/log4j-api-test-2.25.1.jar
[INFO] 
[INFO] --- cyclonedx:2.9.1:makeAggregateBom (generate-sbom) @ log4j-api-test ---
[INFO] CycloneDX: Resolving Dependencies
[INFO] CycloneDX: Creating BOM version 1.6 with 58 component(s)
[INFO] CycloneDX: Writing and validating BOM (XML): /var/maven/app/log4j-api-test/target/bom.xml
[INFO]            attaching as log4j-api-test-2.25.1-cyclonedx.xml
[INFO] 
[INFO] --- source:3.3.1:jar-no-fork (attach-sources) @ log4j-api-test ---
[INFO] Building jar: /var/maven/app/log4j-api-test/target/log4j-api-test-2.25.1-sources.jar
[INFO] 
[INFO] --- gradle-module-metadata:1.0.1:gmm (default) @ log4j-api-test ---
```
I'd say that in the reference build, the execution order between gradle-module-metadata:1.0.1:gmm and source:3.3.1:jar-no-fork is not the same during the reference build, no idea why

I tested with the deploy profile, it does not change the output, just adds complexity du to use of nexus staging plugin that tricks m-deploy-p (I hate that type of tricks)

hboutemy:
looking at effective pom, it seems there are other extensions used: I fear it introduces internal tricks in the effective model

I'll merge this as-is: nobody seems able to reproduce, it seems it's a fact

ppkarwasz:
> yes, I saw the first and the third issue: for the second, selecting expected Maven version is easy, so not a strong problem (it does not "affect reproducibility" but affects ease of reproducing as rebuild environment is just more strict)

Still, as you know, I prefer to keep the rebuild environment as generic as possible. When I find the time, I really want to resume work on the whole CRLF vs LF nonsense—it’s frankly ridiculous that we're still dealing with this in 2025. I’ll probably start with the CycloneDX Maven Plugin.

> on the third, looking at build log
> 
> [...]
> 
> I'd say that in the reference build, the execution order between gradle-module-metadata:1.0.1:gmm and source:3.3.1:jar-no-fork is not the same during the reference build, no idea why

Did you try rebuilding on a GitHub runner or run the build from `/home/runner/work/logging-log4j2/logging-log4j2` to rule out factors like differing absolute paths?

For reference, this is the original release build we ran:

https://github.com/apache/logging-log4j2/actions/runs/16091508381/job/45409360319

And here's the execution order of the relevant plugins during the `package` phase:

```
[INFO] --- jar:3.4.2:jar (default-jar) @ log4j-api-test ---
[INFO] Building jar: /home/runner/work/logging-log4j2/logging-log4j2/log4j-api-test/target/log4j-api-test-2.25.1.jar
[INFO] 
[INFO] --- gradle-module-metadata:1.0.1:gmm (default) @ log4j-api-test ---
[INFO] 
[INFO] --- cyclonedx:2.9.1:makeAggregateBom (generate-sbom) @ log4j-api-test ---
[INFO] CycloneDX: Resolving Dependencies
[INFO] CycloneDX: Creating BOM version 1.6 with 58 component(s)
[INFO] CycloneDX: Writing and validating BOM (XML): /home/runner/work/logging-log4j2/logging-log4j2/log4j-api-test/target/bom.xml
[INFO]            attaching as log4j-api-test-2.25.1-cyclonedx.xml
[INFO] 
[INFO] --- source:3.3.1:jar-no-fork (attach-sources) @ log4j-api-test ---
[INFO] Building jar: /home/runner/work/logging-log4j2/logging-log4j2/log4j-api-test/target/log4j-api-test-2.25.1-sources.jar
```

Maybe there's a subtle interaction between plugin goals and profile activation that leads to this variance? Let me know if you spot anything suspicious.

hboutemy:
> When I find the time, I really want to resume work on the whole CRLF vs LF nonsense

trying to "fix" what you call "the whole CRLF vs LF nonsense" is nonsense: trying to trick the fact that there are key aspects of environment that are meaningful is wasting time. Some people build on Windows, some people build on *nix, that's a fact you can't fight

> Did you try rebuilding on a GitHub runner or run the build from /home/runner/work/logging-log4j2/logging-log4j2 to rule out factors like differing absolute paths?

I tried to rebuild in many conditions, both GH runner and local

> For reference, this is the original release build we ran:
> https://github.com/apache/logging-log4j2/actions/runs/16091508381/job/45409360319

great, with that precise info on the effective run during the release, we can chase precisely the differences

ppkarwasz:
> > When I find the time, I really want to resume work on the whole CRLF vs LF nonsense
> 
> trying to "fix" what you call "the whole CRLF vs LF nonsense" is nonsense: trying to trick the fact that there are key aspects of environment that are meaningful is wasting time. Some people build on Windows, some people build on *nix, that's a fact you can't fight

These days even Windows works fine with LF as line separator. I don't see any reason why a CycloneDX SBOM could not be generated with UNIX-style lines.

PR #3337, opened 2025-07-31
add plotsquared-bukkit 7.5.6

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/intellectualsites/plotsquared/plotsquared-bukkit/README.md


hboutemy:
Gradle project that uses https://vanniktech.github.io/gradle-maven-publish-plugin/other/

need to adapt the command to deploy to local repository, I don't know how = where Gradle expertise can be useful

PR #3338, opened 2025-07-31
Update Hibernate Search build spec

marko-bekhta:
Hey 🙂 👋🏻 
Here's an update to the Hibernate Search build spec. Current one picks up a pom that is not actually published (and shouldn't be).

- Use newer Maven (though the build is using wrapper, but as the spec file is modified .. let's bump it)
- Remove the now unnecessary disabling skip detection, with the move to Maven Central the nexus plugin is not used anymore, instead, the deploy plugin is used and the artifacts will be correctly detected.


hboutemy:
great, thanks

PR #3356, opened 2025-08-04
Removes UMASK environment variable from Maven settings.xml

### LINKED ISSUE #3135: Maven 4.0.0-rc-4 builds are broken for reproducible central
Please see #3134 as it's failing.
Thank you

froque:
Apache Maven 4.0.0-rc-4 fails with:
> [ERROR] Internal error: java.lang.IllegalArgumentException: Invalid LocalRepositories: [/var/maven/.m2/repository-${MVN_UMASK} ()] -> [Help 1]
org.apache.maven.InternalErrorException: Internal error: java.lang.IllegalArgumentException: Invalid LocalRepositories: [/var/maven/.m2/repository-${MVN_UMASK} ()]
>     Suppressed: java.lang.IllegalArgumentException: Not fully interpolated local repository /var/maven/.m2/repository-${MVN_UMASK} ()

fixes #3135

### DIFF
--- a/m2/settings.xml
+++ b/m2/settings.xml
@@ -1,5 +1,5 @@
 <settings>
-  <localRepository>${user.home}/.m2/repository-${MVN_UMASK}</localRepository>
+  <localRepository>${user.home}/.m2/repository</localRepository>
   <profiles>
     <profile>
       <id>staged-releases</id>


hboutemy:
yes, this would be an option to fix #3135
but it would break buids that rely on this settings: we need to have different local repositories based on umask
need to find another strategy that dropping the feature, or letting `settings.xml` interpolation do the job (as was properly working until now...)

froque:
Any example of a build broken by this ? 

I tried with a few projects and could not find a problem. 

> we need to have different local repositories based on umask

Why ?

hboutemy:
when I added the feature, it was on purpose: cases where project copies dependencies from local repo, then inherit permissions
yes, not the vast majority
it was a long time ago, need to investigate history to find

hboutemy:
thinking at it: perhaps just using `mvn -Dmaven.repo.local=...` can be a good option on the few buildspecs where umask is important for local repo
I need to test

hboutemy:
for reference and further research: this was introduced in ed7fb4f45f9e795647219384f79956ab8640a360

hboutemy:
another aspect coming back to my mind: for the umask from repository issue to be seen, we need 2 builds:
- 1 that fetches the content with a umask
- the other that uses the content from the repository, expecting another umask

these conditions happen when I'm building on my laptop, because local repository is kept between all my builds
but it does not happen on GHA builds, that start from an empty local repo each time

hboutemy:
ok, no time to find a good example from the past
given:
1. nowadays most builds are done on GHA with fresh local repo, that will be filled with the current umask
2. I'll remember when I'll be hit again in the future (because I'll be hit again in the future, I'm sure)
I'll just merge this and rework the umask-specific local repository solution later...

thanks for your patience...

PR #3452, opened 2025-08-15
add mockito-core 5.19.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/mockito/mockito-core/README.md

### DIFF
--- a/content/org/mockito/mockito-core/mockito-5.19.0.buildspec
+++ b/content/org/mockito/mockito-core/mockito-5.19.0.buildspec
@@ -1,0 +1,19 @@
+groupId=org.mockito
+artifactId=mockito-core
+display=${groupId}:${artifactId}
+version=5.19.0
+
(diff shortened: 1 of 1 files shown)


hboutemy:
@JuyeopJang it seems the property from https://github.com/mockito/mockito/issues/3563 is ignored: did I miss something?

JuyeopJang:
Hi @hboutemy,

Sorry, I didn’t fully consider this earlier. The fix in my previous PR works for modules like mockito-core, but for the three modules you mentioned, which use the OSGi Bnd plugin, the Created-By value wasn’t applied. I’m preparing a new PR to address this, so I’d appreciate it if you could wait a bit.

hboutemy:
> I’d appreciate it if you could wait a bit

in fact, Mockito has the lead for the releases: I'm just rebuilding with what is available in the releases :)
so the fix is not in 5.19.0, but hopefully in 5.19.1

PR #3726, opened 2025-09-16
add gson-parent 2.13.2

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/google/code/gson/README.md


hboutemy:
ideally, requires https://github.com/apache/maven-artifact-plugin/issues/173

have to write `buildinfo.skipModules` by hand instead

PR #3765, opened 2025-09-18
add guava 33.5.0-jre

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/google/guava/README.md


hboutemy:
bug in artifact:compare: https://github.com/apache/maven-artifact-plugin/issues/181 for `buildinfo.ignore`

in the current case, `buildinfo.skipModules` looks a better choice

PR #3865, opened 2025-09-27
add mada-style-gradle 1.20

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/dk/mada/style/mada-style-gradle/README.md


jskov:
This was badly tagged. I have just tagged 1.2.0.
I will have a look at the build - it requires Java 25.

jskov:
Dang. I have switched to an atomic OS, and cannot get containers to run.
Not sure when I will get the time to run this in a VM.
So I guess let it lapse until then?
Ta!

hboutemy:
> This was badly tagged. I have just tagged 1.2.0.

FYI, detection comes from existence in Maven Central: https://central.sonatype.com/artifact/dk.mada.style/mada-style-gradle/versions
looking to Git tag in source is only a consequence

IIUC, that means that 1.20 has not public source code any more? or that it's only the Git tag that has been deleted, but Git commit is still there?

jskov:
The commit hash is unchanged - but I deleted the badly named tag (from the repo).
The artifact is still on MC.
As the other issue, it was mostly with a focus on cleanup I would delete this.

PR #3868, opened 2025-09-28
Remove dk-mada-reproducible

jskov:
I have archived the project; it has been obsoleted by Gradle 9.0.0


hboutemy:
I understand that it has been obsoleted, then there is no expected future
but history is history, why dropping info on history?

jskov:
No reason other than keeping things relevant.
If you do not mind, just leave it.
Ta.

PR #3990, opened 2025-10-10
add maven-resolver 2.0.13

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/maven/resolver/maven-resolver/README.md


hboutemy:
why does cyclonedx-maven-plugin download content from the release when the release is not done yet?
strange...
nothing the project can fix at its level
suprising that it happens on this release but not on other projects using this plugin

Issue #4021, opened 2025-10-16
Instructions for container rebuild using podman no longer work on SELinux host


gemmellr:
Following the same instructions from TOOLS.md as many times before, for running a reproducible build check using containers run via podman, I now get an error before much of anything happens:

>  Error: /path/to/buildcache/component:/var/maven/app:Z,rw:z: incorrect volume format, should be [host-dir:]ctr-dir[:option]

Looking at recent changes I believe that #3906 from @ppalaga will have lead to this, since it conflicts with the existing instructions in https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/doc/TOOLS.md#1-rebuild-yourself-to-check-results , resulting in both ":Z,rw" and ":z" flags trying to be applied and resulting an an invalid format.

Specifically, the doc notes to set the following when using podman:
```
# Extra flags to use when mounting volumes in the container. Defaults to "", but for podman running on an SELinux host, you need ":Z,rw"
RB_OCI_VOLUME_FLAGS
```

Realising these conflicting options were the issue, simply unsetting `RB_OCI_VOLUME_FLAGS=":Z,rw"` got the rebuild to progress, as presumably would have leaving the flags and reverting that commit.

At minimum the docs would seem to need updated now, but I also wonder whether the script should be using ":Z,rw" like the original docs suggested, or continue using ":z" as it now does?

ppalaga:
Thanks for the report @gemmellr! 
Could you please check, whether https://github.com/jvm-repo-rebuild/reproducible-central/pull/4058 works for you on podman?

gemmellr:
Just did, looks good, works with both the original usage insturctions now, and the simplified usage it now enables.

PR #4058, opened 2025-10-20
Fixup f88b641 Mount volumes with :z if SELinux is enabled

ppalaga:
Fix https://github.com/jvm-repo-rebuild/reproducible-central/issues/4021


gemmellr:
I checked that this worked for me on podman when trying:
- The simplified case of `export RB_OCI_ENGINE=podman` alone, that these/previous changes now enables to work.
- Specifying both `export RB_OCI_ENGINE=podman` and `export RB_OCI_VOLUME_FLAGS=":Z,rw"` together, as per the original instructions that f88b641 broke.

hboutemy:
thank you

Issue #3949, opened 2025-10-06
Add caffeine cache


ben-manes:
[Caffeine cache](https://github.com/ben-manes/caffeine) has had reproducible builds since v3.1.3 (2023). Admittedly there is no unit test in the repository to assert that, so I'll add that to avoid a regression.

hboutemy:
I just tried to create a buildspec for your project = [97a0b5e](https://github.com/jvm-repo-rebuild/reproducible-central/commit/97a0b5ecc803ff8d280d7c616b81a2c4d5d45bf0) #4063

please help at #4063, because it does not yet work

and I'd love your help too at writing doc from Gradle users to write and contribute their buildspec in the future: I need help, I'm not a Gradle expert

ben-manes:
Thanks for giving it a try! Can you tell me what didn't work and how to debug?

hboutemy:
see #4063 , I just wrote where I need help

hboutemy:
Caffeine Cache is now tracked https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/github/ben-manes/caffeine/README.md
history has been evaluated from 3.2.0 (to current 3.2.2)

given you said it's expected to be Reproducible since 3.1.3, I tried to rebuild 3.1.3 (`bin/add-new-release.sh content/com/github/ben-manes/caffeine/caffeine-cache-3.2.0.buildspec 3.1.3` creates buildspec for 3.1.3 from existing 3.2.0), but the build failed

ben-manes:
Thanks @hboutemy. I thought so but in a fresh review, I see the gradle `isPreserveFileTimestamps` in 3.1.7 as part of the groovy to kotlin migration. I thought it preceded that and I looked when writing that note, but I can't explain why I thought 3.1.3. Here is the [section](https://github.com/ben-manes/caffeine/blob/84dafa0c362f08c4ec9e1af4782fd23bf3eed621/gradle/plugins/src/main/kotlin/lifecycle/java-library-caffeine-conventions.gradle.kts#L46-L55) that should have made it reproducible.

ben-manes:
oh this is the [groovy version](https://github.com/ben-manes/caffeine/blob/b9ad083e93d642bbff388f2f818ef6e1a294646d/gradle/codeQuality.gradle#L70-L73) in 3.1.3. I think it was close but maybe not because it lacked the file permissions.

```groovy
tasks.withType(AbstractArchiveTask) {
[...]

PR #4063, opened 2025-10-21
add Caffeine Cache 3.2.2 #3949

hboutemy:
you can rebuild yourself locally with `./rebuild.sh content/com/github/ben-manes/caffeine/caffeine-cache-3.2.2.buildspec`

as you can see from result, checkout of `v3.2.2` Git tag and running `./gradlew clean assemble -x test -x signMavenPublication --no-daemon --no-build-cache publishToMavenLocal` creates artifacts for `3.2.2-SNAPSHOT`

is there a specific command expected to build `3.2.2`?

### DIFF
--- a/content/com/github/ben-manes/caffeine/README.md
+++ b/content/com/github/ben-manes/caffeine/README.md
@@ -22,6 +22,6 @@ rebuilding **1 releases** of com.github.ben-manes.caffeine:caffeine:
 
 | version | [build spec](/BUILDSPEC.md) | [result](https://reproducible-builds.org/docs/jvm/): reproducible? | [stabilize](https://github.com/google/oss-rebuild/blob/main/cmd/stabilize/README.md) | size |
 | -- | --------- | ------ | ------ | -- |
-| [3.2.2](https://central.sonatype.com/artifact/com.github.ben-manes.caffeine/caffeine/3.2.2/pom) | [gradle jdk17](caffeine-cache-3.2.2.buildspec) | [result](caffeine-3.2.2.buildinfo): [15 :white_check_mark:  1 :warning:](caffeine-3.2.2.buildcompare) | 1 :rotating_light: | 186M |
+| [3.2.2](https://central.sonatype.com/artifact/com.github.ben-manes.caffeine/caffeine/3.2.2/pom) | [gradle jdk17](caffeine-cache-3.2.2.buildspec) | [result](caffeine-3.2.2.buildinfo): [15 :white_check_mark:  1 :warning:](caffeine-3.2.2.buildcompare) [:mag:](caffeine-3.2.2.diffoscope) | 1 :rotating_light: | 186M |
 
 <i>(size is calculated without javadoc, that has been excluded from reproducibility checks)</i>
--- a/content/com/github/ben-manes/caffeine/caffeine-3.2.2.buildinfo
+++ b/content/com/github/ben-manes/caffeine/caffeine-3.2.2.buildinfo
@@ -13,8 +13,8 @@ outputs.1.0.length=731178
 outputs.1.0.checksums.sha512=ad343e9ee3271a54f58dd6f26473a5825d7d007810725d5a8dcb249dd455be7c667a7a0b1e1c2bb2bbdf34c49649ccfae4ee1023ac66a2fbb14609abb8681275
 
 outputs.1.1.filename=caffeine-3.2.2-test.jar
-outputs.1.1.length=1198053
-outputs.1.1.checksums.sha512=d6f997f253b9a6663442e2eaeedffeaa87ef0bd2125a32e97891293a5c628985dfac79491a195982a3718107ecea65e8a575ea7275ccd2e389a257370dd050c5
+outputs.1.1.length=1198052
+outputs.1.1.checksums.sha512=d358e6ccca2244ba5c611fea6b1f458949f1bf63570c854d9e4477e54831e89eff80761a82d48191963236a9a5f9ae7d3ac23dd446cbd352dbc9f4c16fa3a483
 
 outputs.1.2.filename=caffeine-3.2.2.jar
 outputs.1.2.length=895419
--- a/content/com/github/ben-manes/caffeine/caffeine-3.2.2.diffoscope
+++ b/content/com/github/ben-manes/caffeine/caffeine-3.2.2.diffoscope
@@ -1,0 +1,101 @@
+1 / 1 central/com/github/ben-manes/caffeine/caffeine/3.2.2/caffeine-3.2.2-test.jar repository/com/github/ben-manes/caffeine/caffeine/3.2.2/caffeine-3.2.2-test.jar
+--- central/com/github/ben-manes/caffeine/caffeine/3.2.2/caffeine-3.2.2-test.jar
++++ repository/com/github/ben-manes/caffeine/caffeine/3.2.2/caffeine-3.2.2-test.jar
+├── zipinfo {}
+│ @@ -1,8 +1,8 @@
+│ -Zip file size: 1198050 bytes, number of entries: 517
+│ +Zip file size: 1198052 bytes, number of entries: 517
+│  drwxr-xr-x  2.0 unx        0 b- defN 80-Feb-01 00:00 META-INF/
+│  -rw-r--r--  2.0 unx       25 b- defN 80-Feb-01 00:00 META-INF/MANIFEST.MF
+│  drwxr-xr-x  2.0 unx        0 b- defN 80-Feb-01 00:00 com/
+│  drwxr-xr-x  2.0 unx        0 b- defN 80-Feb-01 00:00 com/github/
+│  drwxr-xr-x  2.0 unx        0 b- defN 80-Feb-01 00:00 com/github/benmanes/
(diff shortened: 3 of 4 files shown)


ben-manes:
Oh yes, this is custom to Caffeine where by default it uses -SNAPSHOT and the `-Prelease` flag will drop that.

Gradle has a built-in property for the version which can otherwise be set externally via `-Pversion`. Some projects allow this to be set externally by defining their defaults in `gradle.properties`, but many others set this via other mechanisms like release plugins or explicitly like Caffeine does. I can adjust Caffeine to honor if an external setting in the next release, but currently it is done in the [root build file](https://github.com/ben-manes/caffeine/blob/master/build.gradle.kts).

The most reliable way is to use an [init script](https://docs.gradle.org/current/userguide/init_scripts.html). This way you can enforce your logic regardless of the project's customizations which will make it more universal.

```kts
allprojects {
  afterEvaluate {
    version = (findProperty("forcedVersion") ?: return@afterEvaluate) as Any
  }
}
```

```console
➜  caffeine git:(master) ✗ ./gradlew --init-script reproducible-build.gradle.kts -PforcedVersion=9.9.9 jar --no-build-cache --console colored

Parallel Configuration Cache is an incubating feature.
Configuration on demand is an incubating feature.
Calculating task graph as configuration cache cannot be reused because the set of Gradle properties has changed: the value of 'forcedVersion' was changed.
> Task :plugins:checkKotlinGradlePluginConfigurationErrors SKIPPED
> Task :plugins:generateExternalPluginSpecBuilders UP-TO-DATE
> Task :plugins:extractPrecompiledScriptPluginPlugins UP-TO-DATE
> Task :plugins:compilePluginsBlocks UP-TO-DATE
> Task :plugins:generatePrecompiledScriptPluginAccessors UP-TO-DATE
> Task :plugins:generateScriptPluginAdapters UP-TO-DATE
> Task :plugins:compileKotlin UP-TO-DATE
> Task :plugins:compileJava NO-SOURCE
> Task :plugins:pluginDescriptors UP-TO-DATE
> Task :plugins:processResources UP-TO-DATE
> Task :plugins:classes UP-TO-DATE
> Task :plugins:jar
> Task :caffeine:processResources NO-SOURCE
> Task :guava:downloadCaffeine UP-TO-DATE
> Task :caffeine:processCodeGenResources NO-SOURCE
> Task :jcache:downloadCaffeine UP-TO-DATE
> Task :guava:processResources NO-SOURCE
> Task :caffeine:downloadCaffeine UP-TO-DATE
> Task :jcache:processResources UP-TO-DATE
> Task :simulator:downloadCaffeine UP-TO-DATE
> Task :caffeine:processJavaPoetResources UP-TO-DATE
> Task :simulator:processResources UP-TO-DATE
> Task :caffeine:compileJavaPoetJava
> Task :caffeine:javaPoetClasses
> Task :caffeine:generateNodes UP-TO-DATE
> Task :caffeine:generateLocalCaches UP-TO-DATE
> Task :caffeine:compileJava
> Task :caffeine:classes
> Task :jcache:compileJava
> Task :jcache:classes
> Task :simulator:compileJava
> Task :simulator:classes
> Task :caffeine:compileCodeGenJava
> Task :caffeine:codeGenClasses
> Task :caffeine:jar
> Task :jcache:jar
> Task :guava:compileJava
> Task :guava:classes
> Task :guava:jar
> Task :simulator:jar

[Incubating] Problems report is available at: file:///Users/ben/projects/caffeine/build/reports/problems/problems-report.html

Deprecated Gradle features were used in this build, making it incompatible with Gradle 10.

You can use '--warning-mode all' to show the individual deprecation warnings and determine if they come from your own scripts or plugins.

For more on this, please refer to https://docs.gradle.org/9.2.0-rc-2/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.

BUILD SUCCESSFUL in 40s
28 actionable tasks: 11 executed, 17 up-to-date

Publishing Build Scan to Develocity...
https://caffeine.gradle-enterprise.cloud/s/szc6rqut25mn4

Configuration cache entry stored.
➜  caffeine git:(master) ✗ ls caffeine/build/libs
caffeine-9.9.9.jar
```

hboutemy:
added `-Prelease` to command line, as I understand this part
I don't get the rest

ben-manes:
Thanks! I'll try to explain the rest.

Gradle lacks a versioning and release process in the manner that Maven enforces. It is just an ad hoc project property that is used in jar and metadata where needed. That results in a variety of release plugins (Maven-style, git tag-based, etc), repository publishing plugins, or version setting and incrementing plugins. Gradle lacks a concept of `-SNAPSHOT` which a publishing plugin might use for uploading to the Sonatype snapshot vs Maven Central release repository. So effectively the project must decide how it sets its version and there is not much consistency. Caffeine sets it in the root build file and uses the `release` project property to determine if it should be appended with `-SNAPSHOT` or not, the CI build publishes each commit as a new snapshot, and the CI release publishes with that project property when a Github release tag is cut.

Since there is no consistency across Gradle projects for how to manage versions, you can augment a foreign build to force a version. The `--init-script` provides a build file that you can provide to wrap the external project with your own logic. For example the NullAway static analyzer performs regression testing using external projects where it [forces](https://github.com/uber/NullAway/blob/master/.github/workflows/use-snapshot.gradle.kts) the NullAway dependency to the locally built snapshot. In your case, you can provide an init-script that overrides the project version and use that across all Gradle projects. It could be defined as,

```kts
allprojects {
  afterEvaluate {
    findProperty("forcedVersion").let { version = it as Any }
  }
}
```

This allows you to provide a `forcedVersion` to override the `project.version` property. If the custom property is present then it sets the version property after the project has been evaluated (therefore replacing the original value), so you can use `--init-script reproducible-build.gradle.kts -PforcedVersion=9.9.9` to force any Gradle project to the version string you specify. This lets you ignore the build author's quirks and have consistency across all Gradle projects in reproducible-central.

hboutemy:
IIUC, your init script idea would be a way for Reproducible Central to have a general solution for every project using Gradle?
to avoid the current situation where I have to discover each local project convention

I applied the local convention to check your project: as you can see, 1 artifact is different
I'll commit this short term solution for now

on getting a global solution for every Gradle project, we'll have to sort that out in a separate thread: idea looks nice, but I have more questions

hboutemy:
3.2.2 rebuild result published to https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/github/ben-manes/caffeine/README.md

ben-manes:
Thanks! The unit test jar issue being non-reproducible is very strange. I will try to investigate why 3.2.2 had that problem after work. In master there is [checkBuildReproducibility.sh](https://github.com/ben-manes/caffeine/blob/master/.github/scripts/checkBuildReproducibility.sh), so I'll see if it passes on the prior release tag. It only failed for the `jmhJar`, where [JMH](https://openjdk.org/projects/code-tools/jmh/) is a benchmarking tool and uses bytecode manipulation to instrument classes which makes it non-repeatable and is outside of my control (an unpublished artifact). Gradle 9.0 defaults to reproducible jars and I used to set that explicitly, so I'm unsure what caused the inconsistency on one artifact. Hopefully whatever the cause will be resolved in the next release

hboutemy:
notice: past releases also built
surprisingly, 3.2.0 had no issue

ben-manes:
I let AI grind on it for a while and it determined that it is due to the auto provisioning of the JDK by Gradle's toolchain. This ignores the JDK that you provide and downloads it. In v3.2.0 the JDK was 21, in v3.2.1 it is 24, and on master it is 25. When I release on github actions I provision the JDKs directly and set `-Porg.gradle.java.installations.auto-download=false` so it uses the Temurin build, which allows it to be managed by github's setup-java and utilize its toolchain cache.

---

* Only `MapWithCollisionsProviders$[1-4].class` differ between your jar and the Maven Central artifact; every other entry matches byte-for-byte.
* `cmp` shows a single byte change per class, and `javap -v` reveals it is just the last entry in each `LineNumberTable` mapping bytecode offset 30 to source line 113 locally versus line 112 in the published jar (see `MapWithCollisionsProviders.java`, around the `createCollidingObjectKeys` loop).
* Nothing functional changed—the debug line metadata shifts because the tests were compiled with different JDK builds. Your Gradle run is using the Mandrel-based 24.2 toolchain at `mandrel-java24-24.2.0.0-Final`, while the release pipeline pulls the `Temurin 24+36 HotSpot` bits (see `gradle-daemon-jvm.properties` → Foojay ID → `temurin24-binaries`).
* That explains why 3.2.0 (still compiled with the old toolchain) matches, but 3.2.1/3.2.2 do not.

1. Download the Temurin 24+36 JDK (`OpenJDK24U-jdk_aarch64_mac_hotspot_24_36.tar.gz`) and export `JDK_CI=/path/to/jdk-24+36/Contents/Home` (or set `JAVA_VENDOR=ECLIPSE_ADOPTIUM`) before running `./gradlew clean testJar -Prelease --rerun-tasks`; the resulting SHA-256 will match `e244965101fe431c3b87f5f7623f30c41be53bab200845e12607dc323b280e05`.
2. To avoid future drift, consider pinning the toolchain vendor in `java.toolchain` (e.g. vendor = JvmVendorSpec.ADOPTIUM) or documenting that JDK_CI must point to the Temurin build used for releases.

Once the toolchain is aligned, `LineNumberTable` metadata will be identical and the checksum will match Maven Central.

PR #4183, opened 2025-10-30
add trino-root 478

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/trino/README.md

### DIFF
--- a/content/io/trino/trino-478.buildspec
+++ b/content/io/trino/trino-478.buildspec
@@ -1,0 +1,20 @@
+groupId=io.trino
+artifactId=trino-root
+display=${groupId}:trino
+version=478
+
+gitRepo=https://github.com/trinodb/trino.git
+gitTag=${version}
+
+tool=mvn-3.9.11
+jdk=24
+newline=lf
+timezone="Etc/UTC"
+umask=022
+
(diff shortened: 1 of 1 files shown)


hboutemy:
@wendigo rebuild is ok, but many artifacts were not published to Maven Central
it does not seem it has been configured in pom.xml
do you know what's the logic?
is there a way for Reproducible Central to define a good ignore list?

wendigo:
@hboutemy how we can configure it in the pom?

Cc @martint

hboutemy:
it all depends on what you use to deploy to Maven Central
do you mean that for 478, you did manual cleanup as a first step, but of course would prefer to do something automatic and reliable for future releases?

wendigo:
@hboutemy we are selectively publishing artifacts (basically removing ones that we don't want to from njord artifact store)

wendigo:
@hboutemy ideally we would like to have this codified in our poms but we didn't found a way yet

wendigo:
@hboutemy https://github.com/airlift/airbase/pull/987 I'm working on that :)

hboutemy:
if you use njord, configuring maven-deploy-plugin does the job

PR #4190, opened 2025-10-31
Add autogenerated buildspec for `athena` from OSS Rebuild

algomaster99:
This is an automatically generated buildspec from https://github.com/google/oss-rebuild. There is a [draft PR](https://github.com/google/oss-rebuild/pull/922) which I created to quickly generate this using `go run tools/ctl/ctl.go infer --ecosystem maven --package io.github.gls-athena.starter.data.redis:athena-starter-data-redis   --version 0.0.9 --artifact athena-starter-data-redis-0.0.9.jar  --format=buildspec`.

Eventually, this automatic conversion would be done on Reproducible Central. We can add a CI job that generates this buildspec and submits a PR to RC.


hboutemy:
first reaction: not building the whole project https://github.com/gls-athena/athena-parent is strange for Reproducible Central

algomaster99:
The thing strange for me is why the stabilization does not produce bit-by-bit identical jars. The diffoscope output in the CI job log seems to have a bug as it only contains:
```
1 / 1 target/reference/io.github.gls-athena.starter.data.redis/athena-starter-data-redis-0.0.9.jar target/athena-starter-data-redis-0.0.9.jar
```
and not the actual diff.

hboutemy:
the rebuild script and all other tools have very limited support for "mvn -f <path>", as it is an edge case for building projects (I'm sure we have less than 10 project in Reproducible Central that do not build from root of the source Git repository, a good "find | grep" could probably give the exact real value)

algomaster99:
> rebuild script and all other tools have very limited support for "mvn -f ",

So is that why we might be getting a diff in the upstream and rebuild artifact? When I run locally, I also don't have the upstream artifact under `target/reference` or under `athena-project/athena-starter/athena-starter-data-redis/target/reference`.

PR #4261, opened 2025-11-07
Add Jakarta Validation

marko-bekhta:
hey 🙂 👋🏻 

4.0.0M1 should be almost reproducible... there is some wiered file permisson difference on the sources jar when compared with the published version, but I'm not yet entirely sure where that comes from  😖. Maybe I'll figure it out for the next M2 🙂 🤞🏻

### DIFF
--- a/content/jakarta/validation/jakarta.validation-api/README.md
+++ b/content/jakarta/validation/jakarta.validation-api/README.md
@@ -1,0 +1,18 @@
+[jakarta.validation:jakarta.validation-api](https://central.sonatype.com/artifact/jakarta.validation/jakarta.validation-api/versions) RB check
+=======
+
+[![Reproducible Builds](https://reproducible-builds.org/images/logos/rb.svg) an independently-verifiable path from source to binary code](https://reproducible-builds.org/)
+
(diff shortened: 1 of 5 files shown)


hboutemy:
the small typo commit is hiding the rebuild of the buildspec: please remove it or squash

marko-bekhta:
ahh my bad ... sorry 🫣 squashed it all 🤞🏻

hboutemy:
requires `umask=022` 

in the future, upgrading maven-source-plugin to 3.3.1 minimum will avoid requiring this config https://issues.apache.org/jira/browse/MSOURCES-137

marko-bekhta:
> in the future, upgrading maven-source-plugin to 3.3.1 minimum will avoid requiring this config https://issues.apache.org/jira/browse/MSOURCES-137

nice 👍🏻 I'll go make sure the version is updated 🙂

hboutemy:
I just added the magic `execBefore="find . -type d -exec chmod g+s {} \;"` that was missing to be able to reproduce your build :)

marko-bekhta:
Thanks!

> execBefore="find .....

ah that's a nice trick 😉🙂

I've been looking into where that permission comes from, since it's not there if you just clone the repo, and it seems it's from CI.

BTW... do you think it could be a reasonable improvement to add something to the maven sources plugin to allow "overriding" the packaged sources permissions ?

PR #4293, opened 2025-11-08
add codeowners-parent 1.12.2

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/nl/basjes/codeowners/codeowners-parent/README.md


hboutemy:
require jdk 26

PR #4390, opened 2025-11-15
add codeowners-parent 1.13.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/nl/basjes/codeowners/codeowners-parent/README.md


hboutemy:
require JDK 26 EA, that is not even available in sdkman for now

PR #4403, opened 2025-11-19
add flowlogix 11

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/flowlogix/README.md

### DIFF
--- a/content/com/flowlogix/flowlogix-11.buildspec
+++ b/content/com/flowlogix/flowlogix-11.buildspec
@@ -1,0 +1,19 @@
+groupId=com.flowlogix
+artifactId=flowlogix
+version=11
+
+display=${groupId}:${artifactId}
+
+gitRepo=https://github.com/flowlogix/${artifactId}.git
+gitTag=Version-${version}
+
+tool=mvn-4.0.0-rc-4
+jdk=25
+newline=lf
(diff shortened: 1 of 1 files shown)


lprimak:
Looks like this is failing due to settings.xml being overridden. Does rebuilder stop `.mvn/settings.xml` from being read by maven 4 somehow?
The example module (which is not even deployed to Central) is the one which is failing and it uses custom settings.xml to bring a different repository that has more up-to-date dependencies.

Do you see any way to fix this? I can add a profile that can be disabled so examples aren't being built.

hboutemy:
ah, thanks for the hint: I did not have time to investigate
it's not Maven 4 that overrides, but Reproducible Central rebuild script, to use https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/m2/settings.xml

the optional build of example module could be an option, yes

lprimak:
I have made this update in https://github.com/flowlogix/flowlogix/commit/a6cf89421768db7ec57ee93faf845cb8c8a518ea
Will wait a couple of days and if that proves stable, I will release 11.0.1
That version will have `-Dskip-jee-examples` which will fix the problem

Thanks for your help!

PR #4555, opened 2025-12-02
Fix apparently wrong newline=11 in surefire-3.0.0-M9.buildspec

ppalaga:


hboutemy:
not really improving much... :)

hboutemy:
oh, I did not get that I did a typo `11` instead of `lf`
I'll fix in master:no need to rebuild
FI, the algorithm try to recognize `crlf` as it is when we need magic to work as if Windows even if we are on *nix
so does not change the result

ppalaga:
Thanks a lot @hboutemy !

ppalaga:
> I'll fix in master:no need to rebuild

I still see newline=11 in [master](https://github.com/ppalaga/reproducible-central/blob/master/content/org/apache/maven/surefire/surefire-3.0.0-M9.buildspec#L11). It would be nice to get it fixed.

PR #4696, opened 2025-12-16
add trino-root 479

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/trino/README.md


hboutemy:
skip deploy not yet implemented https://github.com/airlift/airbase/pull/987

PR #4870, opened 2026-01-11
add kubernetes-client-project 7.5.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/fabric8/kubernetes-client/README.md


hboutemy:
require a first run to generate boms: https://github.com/fabric8io/kubernetes-client/blob/main/.github/workflows/release.yaml#L66

PR #4898, opened 2026-01-15
maven-metadata.xml fetching should not succeed in case of HTTP 404

### LINKED ISSUE #111: rebuild.sh should fail on bad maven-metadata.xml
If the wrong group/artifact is provided then it is not possible to download the `maven-metadata.xml`.

If this happens the stored file contains something like this and the process does not fail:
```
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx</center>
</body>
</html>
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
<!-- a padding to disable MSIE and Chrome friendly error page -->
```

In a normal correct situation it contains both the desired groupId and artifactId.
```
<?xml version="1.0" encoding="UTF-8"?>
<metadata>
  <groupId>nl.basjes.parse.httpdlog</groupId>
  <artifactId>httpdlog-parser</artifactId>
  <versioning>
    <latest>5.9.0</latest>
    <release>5.9.0</release>
    <versions>
      <version>1.3.1</version>
      <version>1.4</version>
...
```

Proposal:
After retrieving the `maven-metadata.xml` the resulting file is checked if it contains both

    "<groupId>${groupId}</groupId>"
    "<artifactId>${artifactId}</artifactId>"

if not then the rebuild fails.

Godin:
While working on #4895 I stepped on the problem https://github.com/jvm-repo-rebuild/reproducible-central/pull/4895#discussion_r2691082737 that seems to be described in #111

Note that while `--fail` curl option is not available in old curl versions, it is already used in other places - for example
https://github.com/jvm-repo-rebuild/reproducible-central/blob/cf7a583850aad684bd6ae2955f40da2961bf19a2/bin/wip.sh#L19-L21

---

Closes #111

### DIFF
--- a/bin/includes/fetchSource.sh
+++ b/bin/includes/fetchSource.sh
@@ -23,7 +23,7 @@ then
   info "Fetching maven-metadata.xml from Maven Central"
   metaUrl="https://repo.maven.apache.org/maven2/$(echo ${groupId} | tr '.' '/')/${artifactId}/maven-metadata.xml"
   echo -e "downloading \033[1m${metaUrl}\033[0m to $(pwd)"
-  runcommand curl -s $metaUrl --output maven-metadata.xml || fatal "failed to download maven-metadata.xml"
+  runcommand curl -s --fail $metaUrl --output maven-metadata.xml || fatal "failed to download maven-metadata.xml"
   head -15 maven-metadata.xml
 fi
 


hboutemy:
good catch, thank you

PR #4899, opened 2026-01-15
Yauaa 8.0.0

nielsbasjes:
Yauaa 8 has different JDK needs than the previous versions.
I also found a typo in a variable name and fixed that too.


nielsbasjes:
@hboutemy How can I make the missing files be ignored? These are all files that are intentionally not pushed to maven central.

hboutemy:
you can configure `-Dbuildinfo.skipModules=*/yauaa-devtools-sortlines,*/yauaa-examples-parent,*/yauaa-example-graalvm-native`
https://maven.apache.org/plugins/maven-artifact-plugin/compare-mojo.html#skipModules

nielsbasjes:
Thanks for the info. I'm going to add that to this merge request.

nielsbasjes:
I just realized why I hadn't picked up this mistake of mine.

During the build I do an install (to test reproducibility) which places **all** artifacts in my local ~/.m2/repository instead of just the ones that are eventually deployed. Deploying works for me using the Sonatype provided central-publishing-maven-plugin which has it's own exclude list.

hboutemy:
for the rebuild to happen, .buildspec file has to be in the last commit of the PR (small limitation of the automation)

PR #4938, opened 2026-01-17
docs: Link to non reproducibility issue for Yauaa 7.32.0

nielsbasjes:
The badge.json and README.md for Yauaa were not updated correctly after releasing Yauaa 8.0.0 somehow.
I think that adding this link (which should have been there anyway) will retrigger this generation process.


hboutemy:
see explanations in #4939 that is exactly in the same case as you

now people are interested in RB results, you are too impatient :)

nielsbasjes:
Yes I'm very impatient ;) (Actually; I thought that something had gone wrong.)

Thanks for the explanations.

hboutemy:
as expected, batch added the 8.0.0 result
if this PR still remains useful, please rebase it

hboutemy:
ok, merging

PR #4939, opened 2026-01-17
add openapi-maven-plugin 0.0.29

kbuntrock:
Add version 0.0.29 of the openapi-maven-plugin.

### DIFF
--- a/content/io/github/kbuntrock/openapi-maven-plugin/openapi-maven-plugin-0.0.29.buildcompare
+++ b/content/io/github/kbuntrock/openapi-maven-plugin/openapi-maven-plugin-0.0.29.buildcompare
@@ -1,0 +1,9 @@
+version=0.0.29
+ok=3
+ko=0
+ignored=0
+okFiles="openapi-maven-plugin-0.0.29.pom openapi-maven-plugin-0.0.29.jar openapi-maven-plugin-0.0.29-sources.jar"
+koFiles=""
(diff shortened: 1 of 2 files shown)


hboutemy:
thank you @kbuntrock 
notice that you don't need to provide the PR: it will be generated automatically after the new version is detected by daily batch

https://github.com/jvm-repo-rebuild/reproducible-central/actions/workflows/update-reports.yml

hboutemy:
I just even saw that the batch is necessary to update `maven-metadata.xml` that drives listing of versions: then even if PR is currently merged with the rebuild, it will appear in the report only after the batch, in 8 hours :)

kbuntrock:
> thank you @kbuntrock notice that you don't need to provide the PR: it will be generated automatically after the new version is detected by daily batch
> 
> https://github.com/jvm-repo-rebuild/reproducible-central/actions/workflows/update-reports.yml

Oh, nice, I was not aware of this. 😊 And what is the preferred way in case the `buildspec` file is known to be different from a previous build? (aside from the version). Is it better to provide a PR like this or to work from the result of the batch?


> I just even saw that the batch is necessary to update `maven-metadata.xml` that drives listing of versions: then even if PR is currently merged with the rebuild, it will appear in the report only after the batch, in 8 hours :)

That's alright, I’m sure the millions of users of my project managed to contain their impatience! 😄

PR #4947, opened 2026-01-19
add avaje-http-parent 3.7

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/avaje/http/README.md


hboutemy:
same issue as https://github.com/avaje/avaje-spi-service/issues/78

PR #4974, opened 2026-01-19
add env-config 2.2.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/com/github/sitture/env-config/README.md


hboutemy:
wrong Git tag = points to pom.xml still at 2.1.0, missing GIt commit updating to 2.2.0 https://github.com/sitture/env-config/issues/339

Issue #4933, opened 2026-01-16
How to handle Artemis changing Maven groupId?


gemmellr:
Currently the existing ActiveMQ Artemis release reproducibility has been tracked [here](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/README.md).

However [Artemis has moved to being an independent Apache project](https://artemis.apache.org/news/artemis-tlp), and as part of this is switching Maven groupId from org.apache.activemq to org.apache.artemis.

The first candidate for an Apache Artemis 2.50.0 release under org.apache.artemis groupId is being prepared just now. I have locally prepared a new buildspec under content/org/apache/artemis [1] _(also placing a copy of the staged maven-metadata.xml file [2] to let the rebuild scripts bootstrap, since the Maven Central one they look for otherwise isnt available yet)_ and was able to successfully reproduce the staged new build (prepared by someone else) via the rebuild containers using podman.

However, relocation poms are also being staged/deployed separately at the old org.apache.activemq GAV to alert and aid users switching, and those will be picked up by the existing reproducibility nightly release summary checks, and indeed already were in
[last nights update](https://github.com/jvm-repo-rebuild/reproducible-central/commit/ccb0290e860864116098a2b9d245a005c937f2d7#diff-3a0056cd7af80b81f9f282d730ec91004f032633504606a4e3f1bcc699517c70).

I wonder how you want this to be handled going forward? My reading of [bin/update-reproducibility-summary.sh](https://github.com/jvm-repo-rebuild/reproducible-central/blob/049fa32cdf3dd862bcec809e9d7191a1261380b3/bin/update-reproducibility-summary.sh) suggests that just deleting the old [maven-metadata.xml file](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/maven-metadata.xml) might stop it looking for new ActiveMQ Artemis artifacts both during staging and after release, preventing it looking at the relocation poms, and the checks could then instead just concentrate on the new Apache Artemis releases once a maven-metadata.xml and buildspec is in place for that, starting afresh with Apache Artemis 2.50.0 under content/org/apache/artemis. Would that be the favoured approach? Or possibly better removing the old ActiveMQ Artemis content also so that the main content README only lists 1 entry for Apache Artemis?

[1] Diff of adding local buildspec used to check staged new build:
```
[...]

PR #5036, opened 2026-01-23
add artemis-project 2.50.0

gemmellr:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/artemis/README.md


gemmellr:
Adds initial buildspec for Apache Artemis 2.50.0, and initial maven-metadata.xml file from https://repo1.maven.org/maven2/org/apache/artemis/artemis-project/maven-metadata.xml to bootstrap the update process.

gemmellr:
There are also relocations being published at the old org.apache.activemq groupId for a period going forward, likely until the next major later in the year, that the update process will still see and interpret as new 'ActiveMQ Artemis' releases to be added (#5042). I made [a suggestion](https://github.com/jvm-repo-rebuild/reproducible-central/issues/4933#issuecomment-3773061740) on #4933 around suppressing updates for the old [ActiveMQ Artemis readme](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/README.md) from picking up these relocations, whilst still leaving the previously rebuilt versions being counted and processed for any readme updates from adjustments made in the summary script.

hboutemy:
FYI, only buildspec was necessary in the initial commit, not the maven-metadata. But not a problem :)

the new "ActiveMQ Artemis" 2.50.0 release is a full release, not limited to POMs: checking it is still useful, even if I know that relocation info in the poms will avoid build tools to really use it (if they properly implement relocation handling): not a real problem

thanks for the help: now, you're covered at Reproducible Central regarding this relocation

gemmellr:
> FYI, only buildspec was necessary in the initial commit, not the maven-metadata. But not a problem :)
> 

I guess I didnt look at the PR build enough hehe. I was thinking to the nightly updates, which only run if there is a metadata file, and knew the project summary update script fails if there isnt a metadata so it seemed like it was needed. Didnt spot that the PR build has a 'shell' section that is just descriptive and wont actually run, and it instead updates/creates the metdata file before running the summary in the non-interactive route.

> the new "ActiveMQ Artemis" 2.50.0 release is a full release, not limited to POMs: checking it is still useful, even if I know that relocation info in the poms will avoid build tools to really use it (if they properly implement relocation handling): not a real problem
> 

I dont really see it like that personally. The only thing published into the old ActiveMQ groupId were the stripped down relocation poms that I created, so anything that doesnt implement relocation handling will not find what its looking for at the old groupId as it just isnt there.

I guess my main issue with leaving it this way is that the results of the other/older rebuild (#5042) are not actually even about those relocations either way. They are about exactly the same new Apache Artemis _org.apache.artemis_ artifacts as this build covered, because it runs the same build command on the same source from the same tag. That means it doesnt inspect the relocation poms because they arent deployed by that build (avoiding deploy into two different groupId at the same time, they are published separately). There isnt a single mention of activemq in those 'ActiveMQ Artemis 2.50.0' [buildinfo results](https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/activemq/artemis/artemis-project-2.50.0.buildinfo) as a consequence of that. Its essentially a duplicate build to this one, the resulting buildinfo is the same. Possibly the only difference during that build will be that when the metadata get updated in the other build, it pulls the metadata file from a now-relocation org.apache.activemq pom instead since thats what the buildspec originally referenced, but the versions numbers will still align with this builds.

PR #5044, opened 2026-01-23
add qpid-jms-parent 2.10.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/qpid/jms/README.md


gemmellr:
The buildspec will need updated to use Maven 3.9.x (similarly for qpid-jms 1.16.0, if it is rebuilt after this one)

Since the apache parent pom 37 requires 3.9.x, probably several Apache projects rebuilds could hit this going forward as they update too.

PR #5474, opened 2026-02-28
add byte-buddy-parent 1.18.6-jdk6-jdk5

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/net/bytebuddy/README.md


hboutemy:
looks like failed release, removed from GH, but cannot be removed from Maven Central given its immutability policy
=> will have to wait for 1.18.7 release
https://github.com/raphw/byte-buddy

PR #5806, opened 2026-04-13
Switch to JDK 25 for building Hibernate Models

marko-bekhta:
Hey 👋🏻 🙂 

I noticed the failures for recent releases ... with version 1.1 the build switched to JDK 25 and Gradle 9.x, let's see if the re-builds agree 🤞🏻


hboutemy:
brilliant!
thank you

PR #6112, opened 2026-05-18
add bitcoinj-core-0.17.1

msgilligan:
This fails because `tool=gradle` results in Gradle 9 being used.

This should build successfully with Gradle 8, but the reference build uses the Debian build of Gradle 4.4.x:

https://github.com/bitcoinj/bitcoinj/tree/v0.17.1#building-the-reference-build

One reason we use the Debian build of Gradle is that it is built from source.

The commented-out `execBefore` symbol was the start of an attempt to get the build working with Gradle 9, but additional patches would be required.

### DIFF
--- a/bin/includes/rebuildToolGradle.sh
+++ b/bin/includes/rebuildToolGradle.sh
@@ -26,7 +26,7 @@ rebuildToolGradle() {
 #      jdkImage="openjdk:17-slim"
 #      ;;
 #  esac
-  jdkImage="docker.io/library/gradle:9-jdk${jdk}"
+  jdkImage="docker.io/library/gradle:8-jdk${jdk}"
 
   info "Rebuilding using container image ${jdkImage}"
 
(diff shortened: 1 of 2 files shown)


msgilligan:
The failure is:

```
> Could not set unknown property 'dirMode' for task ':bitcoinj-core:jar' of type org.gradle.api.tasks.bundling.Jar.
```

This setting (and others) are necessary for reproducible builds on older versions of Gradle. `dirMode` and `fileMode` give a warning in Gradle 8 and are removed in Gradle 9.

hboutemy:
all Gradle support for Reproducible Central needs love from Gradle experts: I'm not such an expert, this is currently very experimental

on Gradle, I'd expect the command not to use `gradle` but `./gradlew`: the Docker image is not really provided for Gradle itself but for its prerequisites

msgilligan:
Yeah, using the Gradle wrapper is a pretty standard technique, but some projects (such as this one) do not allow committing binaries/executables.

PR #6113, opened 2026-05-19
add secp256k1-jdk-secp-api-0.2.1

msgilligan:
This release is not in Maven central, so the build will fail. This is WIP for the 0.2.2 release which will be published to Maven Central.


hboutemy:
from what I see, it's not only the release that is not in Maven Central, it's the whole project: previous releases were not either

msgilligan:
> from what I see, it's not only the release that is not in Maven Central, it's the whole project: previous releases were not either
> 
> 

It's my project and it's a new project. I just am investigating reproducibility before releasing to Maven Central. Sorry for the distraction.

PR #6144, opened 2026-05-22
add avaje-metrics-parent 9.9-RC8

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/avaje/metrics/README.md


hboutemy:
missing Git commit for now

PR #6455, opened 2026-06-13
Add buildspecs for Clojure 1.12.5

frenchy64:
https://github.com/frenchy64/reproducible-central/blob/clojure/content/org/clojure/README.md

This adds a buildspec for the latest Clojure: https://github.com/clojure/clojure

Most releases have 3 jars:
- clojure-*.jar
- clojure-*-slim.jar
- clojure-*-sources.jar

clojure-*.jar is never reproducible (yet) because it contains AOT compiled Clojure sources. This is a known limitation to the Clojure team, but there's no one issue to track. My hope is that we can start using the build output here to investigate if we can improve reproducibility or implement custom stabilization of compiled Clojure code.

The other two contain classfiles from compiling Java code and raw `.clj` files and are reproducible after normalization.


hboutemy:
thanks for the hard work
given the result, can we just focus on the latest version? what is the interest in adding all the history?

and on this unique version, publishing the diffoscope to explain the issue would be great

once we have that first small init, we'll see if expanding is useful
and how we can work on fixing for future releases = ideally what I expect from the RB Central work

frenchy64:
@hboutemy thanks for your guidance.

> given the result, can we just focus on the latest version?

Yes, the older versions were mostly for curiosity.

> and on this unique version, publishing the diffoscope to explain the issue would be great

I've included the output of `./build_diffoscope.sh content/org/clojure/clojure-1.12.5.buildcompare buildcache/clojure`.

AFAICT:
- clojure-1.12.5.jar has timestamp and jar ordering issues, alongside a huge amount of bytecode incompatibilities mostly due to Clojure's AOT compilation
- clojure-1.12.5-{sources,slim}.jar have timestamp and jar ordering issues, both resolved by stabilization.

hboutemy:
please drop all the non useful commits: 39 commits, it's a lot of noise
there should be maximum 3:
1. buildspec
2. rebuild
3. addition of diffoscope

and on diffoscope, please don't commit the full file, the Git repo will explode
this is intended for a small analysis, with partial content

eventually: just provide a buildspec (that represents your best rebuild recipe), I'll do the rest

frenchy64:
Done! I reduced the diffoscope file to demonstrate the zip timestamp differences.

hboutemy:
great job: merging

PR #6461, opened 2026-06-14
add org.clojure/test.check 1.1.3

frenchy64:
https://github.com/frenchy64/reproducible-central/blob/test-check/content/org/clojure/test.check/README.md#project-orgclojuretestcheck-

test.check is a source-only Clojure library. No attempt at reproducibility has been made yet.

According to diffoscope, zip order and timestamps are the main non-reproducible elements. Here's the pom.xml: https://github.com/clojure/test.check/blob/master/pom.xml

I'm happy to take suggestions on how to address this, I can follow up with the project to add reproducibility.

### DIFF
--- a/content/org/clojure/test.check/README.md
+++ b/content/org/clojure/test.check/README.md
@@ -1,0 +1,19 @@
+[org.clojure:test.check](https://central.sonatype.com/artifact/org.clojure/test.check/versions) RB check
+=======
+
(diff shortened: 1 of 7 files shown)


hboutemy:
great: with one unique release, I can dive into it

buildspec should also build the source jar: is there a profile that was forgotten?

for next release, you should just need a property as described in https://maven.apache.org/guides/mini/guide-reproducible-builds.html

hboutemy:
ah, and you need a separate directory...

frenchy64:
Investigating the missing jar.

frenchy64:
Moved to new dir, added missing jar.

PR #6552, opened 2026-06-20
project 6.9.0

hboutemy:
- **add jline 4.0.12**
- **add project 6.9.0**


hboutemy:
use Eclipse CBI jarsign? https://github.com/eclipse-ee4j/openmq/blob/master/pom.xml#L369

PR #6908, opened 2026-07-23
Switch to JDK 25 for building Hibernate Reactive

marko-bekhta:
Hello 🙂 👋🏻 
we've switched to building releases with JDK 25 in Hibernate Reactive ... let's see if this also helps with the failed check 🤞🏻


hboutemy:
great finding: I definitively don't understand Gradle :) 

thank you

PR #7028, opened 2026-08-07
add hop 2.17.0

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/apache/hop/README.md


hboutemy:
no idea why, the `source-release.zip` and `-cyclonedx.json` files have not been published, but their signatures are there:
https://repo1.maven.org/maven2/org/apache/hop/hop/2.17.0/ which exactly the same as at Apache https://repository.apache.org/content/repositories/releases/org/apache/hop/hop/2.17.0/

notice: not the source release published at Apache https://archive.apache.org/dist/hop/2.17.0/
it seems the 2 official files are not part of what is tested in Reproducible Central, as they are not published to Central...

PR #7046, opened 2026-08-08
add swagger-ui 5.32.11

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/webjars/swagger-ui/README.md


hboutemy:
not built any more from this Git repository, but through shared WebJARs process (undefined)

PR #7171, opened 2026-08-17
Starting with MCS 0.10.2, use Maven Wrapper to verify the build

mthmulders:


hboutemy:
we don't really care: there are multiple ways to build with the same output

mthmulders:
> there are multiple ways to build with the same output

Correct. However, my line of thought was: by making it explicit that Maven Wrapper should be used, I don't have to update the `tool` whenever the Maven version, that I use for building, updates. Wrapper will just follow along.

hboutemy:
I let you play, if you want, not worth fighting: we'll see if you always use `mvnw` really during your releases in the future, or you'll forget
I saw that in so many cases...

notice: the only significant time when it will be useful is when you switch to Maven 4

mthmulders:
Thanks for merging it! I'm not sure I'll always remember, too. In fact, it might have been (one of the) causes of 0.10.1 being non-reproducible.

hboutemy:
if your build really depends on Maven version, it means you did not pin some plugins versions: you should fix it

for anything else, the only Maven 3 vs Maven 4 aspect counts

mthmulders:
> if your build really depends on Maven version, it means you did not pin some plugins versions: you should fix it

Totally agreed. In that sense, you could argue enforcing the Wrapper duplicates the checks we do in Maven 4 (pinning of plugin versions).

PR #7182, opened 2026-08-17
add hk2-parent 4.0.2

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/org/glassfish/hk2/README.md


hboutemy:
missing Git tag, and even Git commit for this 4.0.2 release

PR #7227, opened 2026-08-20
add jena-kafka 3.2.1

hboutemy:
https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/telicent/jena/jena-kafka/README.md


hboutemy:
Git tag 3.2.1 now exists, but points to 3.2.0

PR #7427, opened 2026-09-12
add io.github.teams4j:teams4j 0.1.0

sh-cho:
New project: teams4j, Adaptive Cards and Microsoft Teams for the JVM (Gradle, 14 modules).

Rebuilt with `gradle:9-jdk21`: 40/40 files match Central.

`kotlinSourcesJar` is excluded from the command. In 0.1.0 it writes the same `-sources.jar` as `sourcesJar` with a different layout, and under parallel execution the later task wins, so a rebuild could differ from Central. The project disables that task from the next release.

https://github.com/jvm-repo-rebuild/reproducible-central/blob/master/content/io/github/teams4j/teams4j-cards/README.md


hboutemy:
perfect, thank you

